PR #18, opened 2022-11-16
Update code gen verification

loosebazooka:
- Updates Makefile to actually generate all required files (-f only includes the last entry)
- Update github check action to make sure we are checking all code

### DIFF
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -13,32 +13,38 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: Generate code
+name: Check generated code for changes
 
 permissions:
   contents: read
 
 on:
   push:
-    paths: 
+    paths:
       - '**.proto'
       - 'gen/**'
   pull_request:
-    paths: 
+    paths:
       - '**.proto'
       - 'gen/**'
 
 jobs:
-  generate_go:
-    name: Generate go code
+  check_generated_protos:
+    name: Check generated protobufs
     runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3
 
-    - name: Compile protobuf
+    # clear out the generated files directory so we know we are actually
+    # generating all the files again (instead of a subset)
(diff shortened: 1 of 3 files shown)


.github/workflows/generate.yml
```diff
@@ -36,9 +36,13 @@ jobs:
     steps:
     - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3
 
-    - name: Compile protobuf
+    - name: Clear out protobuf directory
       run: |
-        make proto
+        rm -rf gen
```
znewman01, line 41: nit: can we make this a `make clean` target?

znewman01 (commented):
Does it actually make sense to commit the generated code?

We have to do this for Go because Go doesn't have support for generated code, whereas in Java you can generate it as part of the JAR or whatever.

We're going to want to publish this as a JAR too, right? (Okay if that's a different PR)

.github/workflows/generate.yml
loosebazooka: yeah I'm doing that... trying to figure out how to get docker to run as the user so clean will work with sudo on local machines

loosebazooka: all the files are generated with ownership `root:root`

loosebazooka:
> We're going to want to publish this as a JAR too, right? (Okay if that's a different PR)

Yeah that's a bit more involved. Will be in a followup

.github/workflows/generate.yml
znewman01: :shipit:

loosebazooka:
> Does it actually make sense to commit the generated code?

soooo.. yes we can generate the java stuff on the fly with a gradle build file. Lemme just remove all the java stuff from here and keep the fixes to the go gen.

asraa (commented):
Thanks for catching the `-f` bug!

znewman01 (dismissed):

loosebazooka:
@asraa  @znewman01 merge? I don't have permissions

loosebazooka:
oh wait, need to update README

asraa (approved):
Thank you!

PR #21, opened 2022-11-17
Initial java build

loosebazooka:

Contains a skeleton build for java that converts the protos into java code. This build can produce a `jar` but there is no release config yet.


loosebazooka:
@znewman01 @asraa

java/.gitignore
```diff
@@ -0,0 +1,7 @@
+.gradle
+/build
+
+!gradle-wrapper.jar
```
znewman01, line 4: Can you explain this?

Also having big opaque blobs is a little scary—can we indicate the URL it came from, the SHA where we downloaded it, the version, etc?

znewman01 (commented):
I have no idea how Java works but seems okay?

Can you explain a little bit about how this works in the README?

Also seems like the GH Actions just makes sure that the build is successful. I assume at some point there's going to be release artifacts that get stored and eventually published, yeah? 

Maybe we need to figure out a versioning strategy for this repo too. But can handle that later for sure.

java/.gitignore
loosebazooka: yeah that's actually pretty normal for java projects. I'll add the gradle wrapper validation action to this. But more info here: https://github.com/gradle/wrapper-validation-action

loosebazooka:
> Also seems like the GH Actions just makes sure that the build is successful. I assume at some point there's going to be release artifacts that get stored and eventually published, yeah?

Yeah, this is just setting up the build. Release is a bit more involved. It needs pgp keys and maven central credentials. Either we generate these for a sigstore.dev group or I just use my for now till we can get the right process up for it.

loosebazooka:
> Can you explain a little bit about how this works in the README?

What are you looking for here. It says how to generate a jar. But this is a step to a more complete process. The readme will be filled in over time.

asraa (commented):
I also don't know how Java works well, so would only review the docs/README.md.
@patflynn if you're avail can you give a quick pass?

znewman01:
> What are you looking for here. It says how to generate a jar. But this is a step to a more complete process. The readme will be filled in over time.

This repo is maintained largely by non-Java folks, so it'd be nice to just have a summary of how the package generation works: what commands to run, link to https://github.com/gradle/wrapper-validation-action explain the gradle-wrapper.jar nonsense etc.

patflynn:
LGTM. Thanks for doing this so quickly @loosebazooka !!  Linking to the gradle wrapper for further reading makes sense.

znewman01 (approved):
lgtm

Issue #22, opened 2022-11-18
How to release (and version) this repo


znewman01:
We're going to publish libraries across a lot of ecosystems. Don't want a manual process here. Also want the versions that clients see to make sense/generally align. Somewhat tricky because there are two distinct things we're publishing:

- the "spec" should have a version
- the generated code

Proposal (post-1.0):

- We release package versions for every package at once, on releases tagged `v1.0.3` or similar (we have to follow Go conventions because Go needs you to know it's the most important)
- Spec version changes follow semver:
    - Inconsequential tweak (e.g., docs) => bump patch version
    - New feature => bump minor version
    - Breaking change => bump major version
-  If a spec change OR a change to the generation procedure (e.g., bumping protoc version) would cause a change to the generated code, follow semver as if we had written that code:
    - Inconsequential tweak (e.g., docs) => bump patch version (unless you already have bumped patch/minor/major for the spec)
    - New feature => bump minor version (unless you already have bumped minor/major for the spec)
    - Breaking change => bump major version (unless you already have major for the spec)

Annoyingly, a bump to the protoc version might cause a patch change for one language but be a noop for others...in such cases, we can either (a) do a noop release, or (b) skip that release (we will never publish that version for this language, the releases might go `1.0.3` to `1.0.5`)

CC @loosebazooka

loosebazooka:
I'm not sure I have a way to detect changes in Java since we don't check anything in -- maybe we can compare the build result with the last release (will have to wait till we have an actual release), as the build is configured to be reproducible.

Sometimes projects with multiple artifacts (within java) tend to just do releases across the whole suite since it feels consistent (but not functionally different for many artifacts).

I'm okay just always doing releases on all languages if the proto changes or even if one language requires a build update.

loosebazooka:
Would also right now figure out the pre-GA strat so I can import and use this stuff in sigstore-java.

znewman01:
Ahh, I assume pre-GA strat will be "semver lite" (minor version bumps => breaking change, patch bumps => anything nonbreaking) but no guarantees.

znewman01:
Copying  from #26. @tetsuo-cpp writes>

> This repository is eventually going to contain code to publish packages to each language ecosystem's respective package index. We need to figure out some scheme of keeping these version numbers consistent. So for instance, the v0.0.5 version of this package should be generated from the same set of Protobuf specifications regardless of whether we download it for Python, Java, etc.
> 
[...]

PR #25, opened 2022-11-20
Add Python bindings

### LINKED ISSUE #20: Generate Python bindings?
Similar to the codegen'd Go that's already in the repo: would it be okay to generate and check-in the Python bindings? If so, I/we (ToB) can take a stab at it.

tetsuo-cpp:
Closes #20

@znewman01 @woodruffw @di


python/pyproject.toml
```diff
@@ -0,0 +1,42 @@
...
+
+[project.urls]
+Homepage = "https://pypi.org/project/sigstore-protobuf-specs/"
+Issues = "https://github.com/sigstore/protobuf-specs/issues"
+Source = "https://github.com/sigstore/protobuf-specs"
+
+[project.optional-dependencies]
+dev = [
+  "betterproto[compiler]",
+  "grpcio-tools",
```
tetsuo-cpp, line 41: To get things working, I'm listing `grpcio-tools` as a dependency and using the bundled `protoc` instead of invoking the Docker container.

Before this gets merged in, I can fiddle with this to plug `betterproto` into the Docker container's `protoc`. Or we can avoid the Docker container altogether and use this `protoc` for every language.

python/sigstore_protobuf_specs/dev/sigstore/bundle/v1.py
```diff
@@ -0,0 +1,63 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
+# sources: sigstore_bundle.proto
+# plugin: python-betterproto
+from dataclasses import dataclass
+from typing import List
+
+import betterproto
+
+from .dev.sigstore.common import v1
```
tetsuo-cpp, line 9: Unfortunately these relative imports are busted. I'm trying to figure out whether this is a case of PEBCAK or whether there's a bug that needs fixing in `betterproto`.

Makefile
```diff
@@ -23,6 +23,17 @@ go:
...
+python: env/pyenv.cfg
+	@echo "Generating python protobuf files"
+	. env/bin/activate && cd ./python/sigstore_protobuf_specs && python -m grpc_tools.protoc -I ../../protos/ --python_betterproto_out=. ../../protos/*
+
+env/pyenv.cfg:
+	@echo "Building virtual environment for python protobuf"
+	python3 -m venv env
+	./env/bin/python -m pip install --upgrade pip
+	./env/bin/python -m pip install -e ./python[dev]
+	./env/bin/python -m pip install grpcio-tools
```
tetsuo-cpp, line 35: At the moment, I'm installing `grpcio-tools` to invoke the Protobuf compiler.

Before we merge this, I can either fiddle with this to plug `betterproto` into the Docker image's compiler or we can remove the Docker container altogether and make everything use this one.

gen/pb-python/pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+requires-python = ">=3.7"
+
+[project.urls]
+Homepage = "https://pypi.org/project/sigstore-protobuf-specs/"
+Issues = "https://github.com/sigstore/protobuf-specs/issues"
+Source = "https://github.com/sigstore/protobuf-specs"
+
+[project.optional-dependencies]
+dev = [
+  "betterproto[compiler]==2.0.0b5",
```
tetsuo-cpp, line 40: The developers of `betterproto` have been working on a v2.0.0 release for a few years now. We pretty much have to use a beta version as the current stable version seems to have bugs with relative imports and lacks support for `optional`.

tetsuo-cpp:
Since the specs aren't finalised yet, I haven't added a release workflow to build wheels, publish to PyPI, etc. But that should be pretty trivial to do when we need it.

Makefile
woodruffw: I'm personally okay with that, although IMO the `grpcio-tools` dependency should go in the `[dev]` extra rather than being a standalone dependency here.

gen/pb-python/pyproject.toml
woodruffw: I'm okay with a beta version here, as long as we keep it pinned (which we are).

woodruffw:
> Since the specs aren't finalised yet, I haven't added a release workflow to build wheels, publish to PyPI, etc. But that should be pretty trivial to do in a follow up PR when we need it.

Makes sense! IMO we should also get a skeleton test suite in here before doing an initial release.

Makefile
```diff
@@ -23,6 +23,17 @@ go:
 	@echo "Generating go protobuf files"
 	docker run --pull always --platform linux/amd64 -v ${PWD}:/defs ${PROTOC_IMAGE} -d protos -l go --go-module-prefix github.com/sigstore/protobuf-specs/gen/pb-go
 
+python: env/pyenv.cfg
```
znewman01, line 26: Can you add `python` to the `all:` target?

gen/pb-python/pyproject.toml
```diff
@@ -0,0 +1,41 @@
+[build-system]
+requires = ["flit_core >=3.2,<4"]
+build-backend = "flit_core.buildapi"
+
+[project]
+name = "sigstore-protobuf-specs"
+version = "0.0.1"
```
znewman01, line 7: Note to self (see https://github.com/sigstore/protobuf-specs/issues/22): at some point we're going to want one canonical source of truth for the version number. Or at least a "bump version script" that will update this.

(You don't need to worry for this PR)

Makefile
```diff
@@ -23,6 +23,17 @@ go:
...
+	@echo "Generating python protobuf files"
+	. env/bin/activate && cd ./gen/pb-python/sigstore_protobuf_specs && python -m grpc_tools.protoc -I ../../../protos/ --python_betterproto_out=. ../../../protos/*
+
+env/pyenv.cfg:
+	@echo "Building virtual environment for python protobuf"
+	python3 -m venv env
+	./env/bin/python -m pip install --upgrade pip
+	./env/bin/python -m pip install -e ./gen/pb-python[dev]
```
znewman01, line 34: These components don't appear to pin the build tool versions.

Makefile
```diff
@@ -23,6 +23,17 @@ go:
 	@echo "Generating go protobuf files"
 	docker run --pull always --platform linux/amd64 -v ${PWD}:/defs ${PROTOC_IMAGE} -d protos -l go --go-module-prefix github.com/sigstore/protobuf-specs/gen/pb-go
 
+python: env/pyenv.cfg
+	@echo "Generating python protobuf files"
+	. env/bin/activate && cd ./gen/pb-python/sigstore_protobuf_specs && python -m grpc_tools.protoc -I ../../../protos/ --python_betterproto_out=. ../../../protos/*
+
+env/pyenv.cfg:
```
znewman01, line 30: Though venvs are definitely a *better* solution for Python than Docker, they're a less consistent one. Ideally, you would be able to clone this repo on a machine without Python installed and still update it with a simple `make all`. Is is possible to move these steps *inside* of a Docker container? (You can even keep them exposed as Make targets to run them directly if you'd like)

znewman01 (commented):
TYVM, this is looking great.

Couple of nits, mostly having to do with "can you do things in strictly worse ways because we have to share this repo?"

gen/pb-python/pyproject.toml
woodruffw: +1 -- in the past I've done this with a top-level `VERSION` file, although the downside to that is that it doesn't play nicely with `pyproject.toml`.

woodruffw: (We could also keep it synced automatically with a GitHub Action).

Makefile
tetsuo-cpp: [ffa2192](https://github.com/sigstore/protobuf-specs/pull/25/commits/ffa2192f5cc476fff26ef0f26f5fec413c0b17a5)

Makefile
tetsuo-cpp: Since we'll be moving to executing this within Docker, I think I'll just use the `protoc` in the image.

gen/pb-python/pyproject.toml
tetsuo-cpp: Created #26 to track this.

Makefile
znewman01: > (The `grpcio-tools` dep should definitely be pinned, and we should do that while adding it to `pyproject.toml` per [#25 (comment)](https://github.com/sigstore/protobuf-specs/pull/25#discussion_r1028272826))

That's the one I really care about.

Makefile
woodruffw: Sounds good!

tetsuo-cpp:
Ok! That was a bit more complicated than expected. I've modified the Python build to use `protoc` within the Docker container and removed the use of virtual environments.

I'll leave a few comments to explain why I've done certain things.

Dockerfile
```diff
@@ -1,3 +1,12 @@
-# This Dockerfile only exists to trigger Dependabot to update this tooling
-# dependency.
-FROM namely/protoc-all@sha256:07f1ba9dbe11f5675e2efc8617c9552217dc4c3eb5ccd108f7c3889878dbae50 # v1.50_0
+FROM namely/protoc-all@sha256:07f1ba9dbe11f5675e2efc8617c9552217dc4c3eb5ccd108f7c3889878dbae50
+
+# Install Python pip.
+RUN set -ex && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends \
+    python3-pip
```
tetsuo-cpp, line 7: Because we're executing the plugin within Docker, I need to install the `betterproto` in the image. For that, I need `pip`.

Therefore, I've had to begin building a new image on top of `namely/protoc-all` since the base image no longer has everything we need. This is covered by the Makefile.

Makefile
```diff
@@ -13,19 +13,38 @@
...
 # generate Go protobuf code
-go:
+go: docker-image
 	@echo "Generating go protobuf files"
-	docker run --pull always --platform linux/amd64 -v ${PWD}:/defs ${PROTOC_IMAGE} -d protos -l go --go-module-prefix github.com/sigstore/protobuf-specs/gen/pb-go
+	docker run --platform linux/amd64 -v ${PWD}:/defs ${PROTOC_IMAGE} -d protos -l go --go-module-prefix github.com/sigstore/protobuf-specs/gen/pb-go
+
+python: docker-image
+	@echo "Generating python protobuf files"
+# we need to manually fix the PYTHONPATH due to: https://github.com/namely/docker-protoc/pull/356
```
tetsuo-cpp, line 28: Ran into an ugly issue which I've got a fix for here: https://github.com/namely/docker-protoc/pull/356

Basically, the buggy PYTHONPATH causes the interpreter to look in the current directory for modules. And since we have a module called `io` at the top level of the `sigstore_protobuf_specs`, the interpreter actually imports that instead the standard library `io` on startup.

znewman01 (approved):
This LGTM, thanks so much!

PR #28, opened 2022-11-23
Added the message type for an environment, which will be suggested as a

kommendorkapten:
construct in the client spec.

#### Summary
During the work on trust root, verification and bundles the concept of an environment has been proposed. This PR shows the construct as a protobuf message.

#### Release Note
N/A

#### Documentation
N/A


protos/sigstore_trustroot.proto
```diff
@@ -84,3 +84,23 @@ message TrustedRoot {
...
+// Environment acts a selector to filter down a trust root to a smaller one.
+// A policy should reference an environment, or embed one that can be used
+// during artifact verification to filter the global trust root into a one
+// only containing the relevant instances. An environment does not take
+// temporality into account when selecting the instances from a trust root.
+// Prior to performing verification the temporal aspect should be further
+// examined to reduce the set of instances.
+message Environment {
+        // The name of the environment.
+        string name = 1;
```
znewman01, line 97: I never know whether to stick the name inside the struct, or just have a map<string, Environment>

We should consider both options based on how we're planning to use this

protos/sigstore_trustroot.proto
```diff
@@ -84,3 +84,23 @@ message TrustedRoot {
...
+        // URIs for artifact signature transparency log to include.
+        repeated string tlog_uris = 2;
```
znewman01, line 99: I might prefer to define dedicated types for each (`repeated TlogIdentifier tlogs = 2;`), for maximum flexibility.

protos/sigstore_trustroot.proto
kommendorkapten: Agreed, but I usually tend to replicate the name into the struct, even if it's duplicated as then the struct on its own has _some_ kind of identifier.

protos/sigstore_trustroot.proto
kommendorkapten: Good point, as an example we actually may want to refer to a transparency log by it's log id, not just it's URI.

kommendorkapten:
To clarify the concept of a policy more, I'm inclined to materialize a message for that too in this PR. cc @joshuagl @znewman01

znewman01:
See also https://docs.google.com/document/d/1wDwiGF3LZ81P5cCTYv-TaAzFWpkPCz4EYcaHVkBPfuM/edit (may need to join sigstore-dev@googlegroups.com for access)

haydentherapper:
@kommendorkapten Do we still need this or can this be closed?

kommendorkapten:
Forgot about this @haydentherapper, I'm closing this now and re-opening when it's ready.

PR #31, opened 2022-11-23
CI: python-build

woodruffw:
This should act as a smoketest for the generated Python that @tetsuo-cpp added with #25.

### DIFF
--- a/.github/workflows/python-build.yml
+++ b/.github/workflows/python-build.yml
@@ -51,8 +51,4 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install .[dev]
 
-          # NOTE: We run the sdist and wheel builds manually, since
-          # our package metadata includes relative paths (README.md and LICENSE)
-          # that are outside of the package root. As a result, our sdist
-          # cannot be used to produce a wheel in isolation.
-          python -m build --sdist --wheel
+          python -m build
--- a/.github/workflows/python-build.yml
+++ b/.github/workflows/python-build.yml
@@ -1,0 +1,58 @@
+#
+# Copyright 2022 The Sigstore Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Check Python build
+
+permissions:
+  contents: read
+
+on:
+  push:
+    paths:
+      - '**.proto'
+      - 'gen/pb-python/**'
+  pull_request:
+    paths:
+      - '**.proto'
+      - 'gen/pb-python/**'
+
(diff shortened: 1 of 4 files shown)


.github/workflows/python-build.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Build
+        run: |
+          cd gen/pb-python
+          python -m venv env && source env/bin/activate
+          python -m pip install --upgrade pip
+          python -m pip install .[dev]
+
+          # NOTE: We run the sdist and wheel builds manually, since
+          # our package metadata includes relative paths (README.md and LICENSE)
+          # that are outside of the package root. As a result, our sdist
```
znewman01, line 56: Ugh, that's annoying. Would symlinking in work?

.github/workflows/python-build.yml
```diff
@@ -0,0 +1,58 @@
...
+          # cannot be used to produce a wheel in isolation.
+          python -m build --sdist --wheel
```
woodruffw, line 58: N.B.: This isn't ideal, but it's the best we can do without adding our own copies of the top-level `README.md` and `LICENSE` to the Python package's directory. But perhaps that's what we want anyways, so that we can specialize the text of the `README.md` for the Python package, specifically?

znewman01 (commented):
IIUC this will eventually turn into the "publish python package" workflow?

.github/workflows/python-build.yml
znewman01: How about symlinks? 

+1 to "eventually we'll want our own README" but the license thing is unfortunate

woodruffw:
> IIUC this will eventually turn into the "publish python package" workflow?

Correct!

znewman01 (dismissed):

woodruffw:
Might want to hold off on merging for a moment, since the issue in https://github.com/sigstore/protobuf-specs/pull/31#discussion_r1030843763 means that the sdist is currently broken:

```console
python -m pip install dist/sigstore-protobuf-specs-0.0.1.tar.gz
Processing ./dist/sigstore-protobuf-specs-0.0.1.tar.gz
ERROR: The tar file (/Users/william/devel/protobuf-specs/gen/pb-python/dist/sigstore-protobuf-specs-0.0.1.tar.gz) has a file (/private/var/folders/zj/hy934vnj5xs68zv6w4b_f6s40000gn/T/pip-req-build-iedc3qzu/../../LICENSE) trying to install outside target directory (/private/var/folders/zj/hy934vnj5xs68zv6w4b_f6s40000gn/T/pip-req-build-iedc3qzu)
```

.github/workflows/python-build.yml
woodruffw: Ugh, forgot that `git` and hard links don't play along...

woodruffw: If my understanding of `git` is right, having two copies of `LICENSE` in the repo isn't the worst thing -- the underlying blob storage should dedupe them. Still not the cleanest, but given that are you okay with the extra file @znewman01?

woodruffw: Oh, missed your last comment. I'll go ahead then 🙂

woodruffw:
DCO got confused by my revert, but I can squash that out if desired. Otherwise should be good to go.

woodruffw:
Should be good to go now! I've removed the hack, which means that we're now transitively testing the correctness of the sdist.

znewman01 (approved):

PR #33, opened 2022-12-07
Add release build for java

loosebazooka:
Triggers release builds (not releases to maven central) of java artifacts when a new tag with format `release/java/v1.2.3` is created. A corresponding github release is generated with jar artifacts AND slsa attestation. ex: https://github.com/loosebazooka/protobuf-specs/releases/tag/untagged-019d132496fa307dce96

TODO: follow up with script to release these artifacts to maven central.

fyi: @woodruffw @znewman01

### DIFF
--- a/.github/workflows/java-build-for-release.yml
+++ b/.github/workflows/java-build-for-release.yml
@@ -1,0 +1,92 @@
+name: Build Java Release
+on:
+  push:
+    tags:        
+      # if you change this pattern, make sure jobs.strip-tag still works
+      - 'release/java/v[0-9]+.[0-9]+.[0-9]+'
+jobs:
+  ci:
+    uses: ./.github/workflows/java-build.yml
+
+  strip-tag:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+    steps:
+      - name: process tag 
+        id: version
+        run: |
+          TAG=${{ github.ref_name }}
+          echo "version=${TAG#"release/java/v"}" >> $GITHUB_OUTPUT
+
+  build:
+    runs-on: ubuntu-latest
+    needs: [ci, strip-tag]
+    outputs:
+      hashes: ${{ steps.hash.outputs.hashes }}
+    steps:
+      - name: checkout tag
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
+        with:
+          ref: "${{ github.ref }}"
+
+      - name: Set up JDK 8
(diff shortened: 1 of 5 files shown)


.github/workflows/java-build-for-release.yml
```diff
@@ -0,0 +1,92 @@
...
+
+  strip-tag:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+    steps:
+      - name: process tag 
+        id: version
+        run: |
+          echo "version=$(echo ${{ github.ref_name }} | cut -c15-)" >> $GITHUB_OUTPUT
```
znewman01, line 19: Can you add a comment like `strip "release/java/v"`, I had to think for a sec about what the number 15 meant.

znewman01: Or even

```
TAG=${{ github.ref_name }}
echo "version=${TAG#"release/java/v"}" >> $GITHUB_OUTPUT
```

could be clearer IMO

.github/workflows/java-build-for-release.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Set up JDK 8
+        uses: actions/setup-java@de1bb2b0c5634f0fc4438d7aa9944e68f9bf86cc # v3.6.0
+        with:
+          java-version: 8
+          distribution: 'temurin'
+
+      - name: Build project
+        run: |
+          # override the version in gradle.properties
```
znewman01, line 41: Can you elaborate a little more on why the `java-build.yml` call isn't sufficient here?

It's because (1) you need to stick `version` in there, and (2) `createReleaseBundle` is a different target from `build`?

znewman01 (approved):
LGTM, some nits/questions

.github/workflows/java-build-for-release.yml
loosebazooka: Yeah injecting the version is the main part. The other option is to check in the version on every update to settings.gradle.kts (which I don't like in a repo like this). 

createReleaseBundle is a custom task which generates output in a way that makes it easier to use the slsa-generator AND sign and release on another machine. I'm not tied to this task, but it's the most familiar mechanism to me for now.

.github/workflows/java-build-for-release.yml
loosebazooka: good call.

znewman01 (approved):

loosebazooka:
You can merge this. It wont trigger anything until we create a tag: `release/java/vX.Y.Z` and even then that wont push to maven central -- that requires post processing of the release (which I will add a script in for)

znewman01:
Sounds good. LMK if you want me to test it out

PR #37, opened 2022-12-13
Updated the bundle to clarify the intent better per #30.

### LINKED ISSUE #30: Clarify naming/types around verification bundle members?
Similar to #29 -- I'm doing some reading through the protobufs now that we're getting close to integrating them into `sigstore-python`, and I wanted to check my understanding of a few things 🙂 

Here's the bundle as it's currently defined:

```protobuf
message Bundle {
        // MUST be application/vnd.dev.sigstore.bundle+json;version=0.1
        // when encoded as JSON.
        string media_type = 1;
        VerificationData verification_data = 2;
        dev.sigstore.common.v1.VerificationMaterial verification_material = 3;
        oneof content {
                dev.sigstore.common.v1.MessageSignature message_signature = 4;
                // A DSSE envelope can contain arbitrary payloads.
                // Verifiers must verify that the payload type is a
                // supported and expected type. This is part of the DSSE
                // protocol which is defined here https://github.com/secure-systems-lab/dsse/blob/master/protocol.md
                io.intoto.Envelope dsse_envelope = 5;
        }
        // Reserved for future additions of artifact types.
        reserved 6 to 50;
}
```

There are a couple members here that I currently find a little confusing (although it's very possible I don't understand them, and understanding them will resolve my confusion!):

* `VerificationData` is a pretty generic type name, and my first assumption was that it's a wrapper type for the input that was signed for. But it's actually a wrapper for a list of transparency log entries, plus materials for timestamp verification.
  * Proposal: either rename this to something like `ExtendedVerificationMaterials` (emphasizing that it's additional materials that can be used to verify additional properties), or maybe decompose it directly into the body of the `Bundle`?
* `MessageSignature` contains two parts: both a hash over the input, *and* a signature over the input. I found this a little confusing originally, since the two are conceptually separate (and it's not clear what role the hash plays in the bundle, since it can just be computed from the input being verified).
  * Proposal: either remove `MessageSignature.message_digest`, or decompose it into the surrounding `Bundle`? If we do the latter, we need to emphasize that checking the hash against the input's computed hash becomes important -- otherwise we might end up in another "confused metadata" circumstance where the user "verifies" a bundle without actually pinning it to the input they're verifying.
* `HashOutput` is a type that contains basically this tuple: `(hash-type, digest)`. However, my understanding of its use (per the CT RFC) is that the only valid hashing algorithm is SHA256 (despite SHA512 being defined). Am I missing a use case where we need this field to be generic over multiple hashing algorithms?

kommendorkapten:
Added SAN type for other name and removed (unused) DNS. When matching a SAN, it's now clear if the verifier should expect a regular expression or match against an exact value.

#### Summary
Fixes #30 

#### Release Note
None

#### Documentation
None

cc: @haydentherapper @woodruffw @znewman01


protos/sigstore_common.proto
```diff
@@ -158,15 +158,22 @@ message X509Certificate {
 }
 
 enum SubjectAlternativeNameType {
-        DNS = 0;
-        Email = 1;
+        SUBJECT_ALTERNATIVE_NAME_TYPE_UNSPECIFIED = 0;
+        EMAIL = 1;
         URI = 2;
+        // OID 1.3.6.1.4.1.57264.1.7
```
woodruffw, line 164: N.B.: Might be worth adding a link to <https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#1361415726417--othername-san> here.

woodruffw (approved):
LGTM!

woodruffw (approved):
LGTM!

(Not sure why this double posted.)

protos/sigstore_common.proto
kommendorkapten: Yeah, good call-out!

woodruffw (approved):

haydentherapper (approved):
thanks!

woodruffw:
@kommendorkapten BTW, feel free to merge this ahead of #36, and I'll take care of deconflicting there 🙂

protos/sigstore_common.proto
```diff
@@ -158,15 +158,24 @@ message X509Certificate {
...
 }
 
 message SubjectAlternativeName {
         SubjectAlternativeNameType type = 1;
-        // A regular expression describing the expected value for the SAN.
-        string value = 2;
+        oneof identity {
+                // A regular expression describing the expected value for
+                // the SAN.
+                string regexp = 2;
```
vlsi, line 175: Is "regexp" portable across implementations?
For instance, what is the intended regexp flavour?

kommendorkapten: This is not expected to be portable right now, I'll create an issue on fixing this as we are thinking about adding more support for regular expressions.

PR #39, opened 2022-12-14
gen, protos: remove unnecessary agility in LogId, HashAlgorithm

### LINKED ISSUE #38: Clarifying/reducing agility in a few more messages
#30 is getting a little long, so I'm filing a separate issue to track some other potential changes to the current messages 🙂 

---

## `LogId`

Here's how `LogId` is currently defined:

```protobuf
// LogId captures the identity of a transparency log.
message LogId {
        oneof id {
                // The unique id of the log, represented as the SHA-256 hash
                // of the log's public key, computed over the DER encoding.
                // https://www.rfc-editor.org/rfc/rfc6962#section-3.2
                bytes key_id = 1;
                // Currently not used but proposed by
                // https://datatracker.ietf.org/doc/rfc9162/
                ObjectIdentifier oid = 2;
        }
}
```

IMO, we probably don't want this agility: the message formats will probably have to change significantly anyways if and when Sigstore moves to CT 2.0, so we should probably limit `LogId` to just the `SHA256(DER(pk))` format that CT 1.0 allows.

## `HashAlgorithm` and `SignatureAlgorithm`

Here's how these are currently defined:

```protobuf
enum HashAlgorithm {
        HASH_ALGORITHM_UNSPECIFIED = 0;
        SHA2_256 = 1;
        SHA2_512 = 2;
}

// Subset of known signature algorithms.
enum SignatureAlgorithm {
        SIGNATURE_ALGORITHM_UNSPECIFIED = 0;
        ECDSA_P256_SHA_256 = 1; // See NIST FIPS 186-4
        ECDSA_P256_HMAC_SHA_256 = 2; // See RFC6979
        ED25519 = 3; // See RFC8032
        RSA_PKCS1V5 = 4; // See RFC8017
        RSA_PSS = 5; // See RFC8017
}
```

I might be wrong about this, but I _believe_ we don't need SHA2-512 or RSA PSS: neither of these is mentioned in the CT RFCs. But it's possible these show up in certificates anyways; cc @haydentherapper for thoughts on these.

woodruffw:
Closes #38.


woodruffw:
(cc @haydentherapper)

haydentherapper (approved):

Issue #40, opened 2022-12-15
Should more messages/fields be marked as required?


woodruffw:
I'm a bit of a protobuf neophyte, so apologies if this is a silly or misguided question 🙂 

To my understanding, every field in a protobuf definition is optional by default (with the `optional` keyword really just being a way to mark that explicitly).

Given that many of our fields are absolutely required (such as the signature and inclusion proof), does it make sense to mark those as required at the wire level?

I see that Fulcio's protobuf definitions use the `(google.api.field_behavior) = REQUIRED` syntax, so is something similar to that appropriate here?

haydentherapper:
https://google.aip.dev/203 - Required should be used in my opinion

znewman01:
Can this be closed?

woodruffw:
Yep!

PR #41, opened 2022-12-15
gen, protos: add `REQUIRED` attributes to a few fields

woodruffw:
See #40.

~~WIP, since there are probably other things that can be marked as required.~~


protos/sigstore_bundle.proto
```diff
@@ -48,12 +50,12 @@ message TimestampVerificationData {
...
-                dev.sigstore.common.v1.X509CertificateChain x509_certificate_chain = 2;
+                dev.sigstore.common.v1.PublicKeyIdentifier public_key = 1 [(google.api.field_behavior) = REQUIRED];
+                dev.sigstore.common.v1.X509CertificateChain x509_certificate_chain = 2 [(google.api.field_behavior) = REQUIRED];
         }
         // This is the inclusion promise and/or proof, where
         // the timestamp is coming from the transparency log.
-        repeated dev.sigstore.rekor.v1.TransparencyLogEntry tlog_entries = 3;
+        repeated dev.sigstore.rekor.v1.TransparencyLogEntry tlog_entries = 3 [(google.api.field_behavior) = REQUIRED];
         // Timestamp verification data, over the artifact's signature.
         TimestampVerificationData timestamp_verification_data = 4;
```
woodruffw, line 60: N.B.: I didn't mark this as required, since I wasn't sure if it was a required part of every Sigstore verification (e.g. `sigstore-python` does not currently do anything with timestamps). But if it should be, let me know 🙂

protos/sigstore_bundle.proto
```diff
@@ -48,12 +50,12 @@ message TimestampVerificationData {
...
         oneof content {
-                dev.sigstore.common.v1.PublicKeyIdentifier public_key = 1;
-                dev.sigstore.common.v1.X509CertificateChain x509_certificate_chain = 2;
+                dev.sigstore.common.v1.PublicKeyIdentifier public_key = 1 [(google.api.field_behavior) = REQUIRED];
+                dev.sigstore.common.v1.X509CertificateChain x509_certificate_chain = 2 [(google.api.field_behavior) = REQUIRED];
         }
         // This is the inclusion promise and/or proof, where
         // the timestamp is coming from the transparency log.
-        repeated dev.sigstore.rekor.v1.TransparencyLogEntry tlog_entries = 3;
+        repeated dev.sigstore.rekor.v1.TransparencyLogEntry tlog_entries = 3 [(google.api.field_behavior) = REQUIRED];
```
haydentherapper, line 58: The tlog isn't technically required, since you might be running in a private environment where you don't use tlogs.

protos/sigstore_bundle.proto
haydentherapper: Agreed! You just need a timestamp from somewhere, it could be this or the tlog entry, but it's hard to represent that with required labels

protos/sigstore_common.proto
```diff
@@ -132,7 +133,7 @@ message PublicKeyIdentifier {
 
 // An ASN1. OBJECT IDENTIFIER
 message ObjectIdentifier {
-        repeated int32 id = 1;
+        repeated int32 id = 1 [(google.api.field_behavior) = REQUIRED];
 }
```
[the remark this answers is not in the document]
haydentherapper: I think that's reasonable. Ideally we'd have field behavior tags to dictate certain expected properties too (for example, require the OID to be of a minimum length, or require the key_id to match a sha-256 digest regex)

protos/sigstore_bundle.proto
woodruffw: Makes sense! Yeah, it sounds like it makes sense to annotate neither, and have the client select one or both (but not none).

haydentherapper (dismissed):

woodruffw:
I made `RFC3161SignedTimestamp.signed_timestamp` required as well, under the same presumption that `RFC3161SignedTimestamp` itself might be optional, but its single constituent field probably shouldn't be 🙂

woodruffw:
I think this is good for another pass! I'm not 100% sure why the Java build(s) are failing -- maybe the hints I added for required fields aren't compatible with that codegen backend?

woodruffw:
Also, it looks like `protoc`'s Python generation isn't aware of these `REQUIRED` hints. I'll look into it some more, but that's not ideal...

haydentherapper (approved):

vlsi:
@woodruffw , is there a reason you used `VerificationMaterial verification_material = 2 [(google.api.field_behavior) = REQUIRED]` rather than `required VerificationMaterial verification_material...`?

I guess `required` could be better supported across the generators rather than `google.api.field_behavior`.

vlsi:
Is there a reason `HashOutput` fields were not marked as `required`?
I guess it makes no sense to have `HashOutput` message without a digest.
I would say, it would be great to make `algorithm` required as well.

```
message HashOutput {
        HashAlgorithm algorithm = 1;
        // This is the raw octets of the message digest as computed by
        // the hash algorithm.
        bytes digest = 2;
}
```

woodruffw:
> I guess `required` could be better supported across the generators rather than `google.api.field_behavior`.

My understanding of protobuf 3 is that `required` and `optional` do not actually do anything during code generation: `optional` is a no-op because all fields are optional by default, and `required` was completely removed after protobuf 2. So generation backends that handle protobuf 3 correctly won't respect the `required` keyword anyways. 

The `field_behavior` annotation is a non-standard (although Google originated) extension, and AFAICT it's the most common way to re-add required fields to a protobuf definition. But codegen backends aren't actually required to support it, unfortunately.

woodruffw:
As for why some fields are marked required and others aren't: it's because I only did one pass over the definitions, and probably missed some. There are almost certainly other things that should be marked with `field_behavior = REQUIRED` before the 1.0.

vlsi:
I see. I filed https://github.com/sigstore/protobuf-specs/pull/64 to add more "required".

PR #49, opened 2022-12-20
remove checksum generation in java bundle

loosebazooka:

turns out this is unnecessary when uploading bundles to maven central as they are generated on the server side, so if you upload a hash, it generates hashes for that hash. hash explosion.


asraa (approved):

loosebazooka:
@znewman01 merge?

znewman01 (approved):

PR #54, opened 2023-01-10
Added a mime type to the trusted root, as that's expected to be seria…

kommendorkapten:

#### Summary
The trusted root message is expected to be serialized to a file, and transferred to a client via e.g. TUF, see https://github.com/sigstore/root-signing/pull/584
For that reason, we should version it similar to how the bundle is versioned.

#### Release Note
N/A

#### Documentation
N/A

cc: @asraa @haydentherapper @znewman01


codysoyland (approved):

asraa (approved):
Thank you for catching this from the other PR!

woodruffw (approved):

PR #63, opened 2023-01-20
protos: disambiguate `canonicalized_body`

### LINKED ISSUE #62: Ambiguity around `TransparencyLogEntry.canonicalized_body`
I noticed this while trying to get `sigstore-python`'s WIP bundle generation verifying against `sigstore-js`: here's what the current `canonicalized_body` docs say:

```
        // The canonicalized Rekor entry body, used for SET verification. This
        // is the same as the body returned by Rekor. It's included here for
        // cases where the client cannot deterministically reconstruct the
        // bundle from the other fields. Clients MUST verify that the signature
        // referenced in the canonicalized_body matches the signature provided
        // in the bundle content.
```

I interpreted this as the _body_ of the response returned by Rekor, without the surrounding `apiVersion`, `kind`, etc. envelope. My reasoning there is that that's the canonicalized body that goes directly into SET verification, without any transformation, while including the envelope means that the user has to remove it and then potentially re-canonicalize. 

`sigstore-js` interprets it the other way: they include and expect the entire enveloped form.

IMO either of these interpretations is reasonable, but there probably needs to be exactly one 🙂

woodruffw:
Closes #62.

Let me know if this is too verbose 😅 

### DIFF
--- a/protos/sigstore_rekor.proto
+++ b/protos/sigstore_rekor.proto
@@ -103,11 +103,18 @@ message TransparencyLogEntry {
         // entry was appended to the log, and that the log has not been
         // altered.
         InclusionProof inclusion_proof = 6;
-        // The canonicalized Rekor entry body, used for SET verification. This
-        // is the same as the body returned by Rekor. It's included here for
-        // cases where the client cannot deterministically reconstruct the
-        // bundle from the other fields. Clients MUST verify that the signature
-        // referenced in the canonicalized_body matches the signature provided
-        // in the bundle content.
+        // The canonicalized transparency log entry, used to reconstruct
+        // the Signed Entry Timestamp (SET) during verification.
+        // The contents of this field as the same as the `body` field in
+        // a Rekor response, meaning that it does **not** include the "full"
+        // canonicalized form (of log index, ID, etc.) which are
+        // exposed as separate fields. The verifier is responsible for
+        // combining the `canonicalized_body`, `log_index`, `log_id`,
+        // and `integrated_time` into the payload that the SET's signature
+        // is generated over.
+        //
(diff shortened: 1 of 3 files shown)


woodruffw:
cc @haydentherapper

asraa (approved):
Thank you for clarifying the client responsibility here! LGTM

Only curious if you want to include an example of one. But I actually think example full bundles should be written i a separate doc away from the codegen.

woodruffw:
> Only curious if you want to include an example of one. But I actually think example full bundles should be written i a separate doc away from the codegen.

Yeah, I think an example would be beneficial, but agreed about putting it somewhere else 🙂

protos/sigstore_rekor.proto
```diff
@@ -103,11 +103,18 @@ message TransparencyLogEntry {
...
         InclusionProof inclusion_proof = 6;
-        // The canonicalized Rekor entry body, used for SET verification. This
-        // is the same as the body returned by Rekor. It's included here for
-        // cases where the client cannot deterministically reconstruct the
-        // bundle from the other fields. Clients MUST verify that the signature
-        // referenced in the canonicalized_body matches the signature provided
-        // in the bundle content.
+        // The canonicalized transparency log entry, used to reconstruct
+        // the Signed Entry Timestamp (SET) during verification.
+        // The contents of this field as the same as the `body` field in
```
haydentherapper, line 108:
```suggestion
        // The contents of this field are the same as the `body` field in
```

haydentherapper (approved):
Thanks!

haydentherapper (approved):

bdehamer (approved):

PR #64, opened 2023-01-22
Add field_behavior=REQUIRED when it makes sense

vlsi:
#### Summary

This is a follow-up to https://github.com/sigstore/protobuf-specs/pull/41

#### Release Note

More protobuf fields are marked with field_behavior=REQUIRED

#### Documentation

NONE


protos/sigstore_bundle.proto
```diff
@@ -55,15 +55,15 @@ message VerificationMaterial {
         }
         // This is the inclusion promise and/or proof, where
         // the timestamp is coming from the transparency log.
-        repeated dev.sigstore.rekor.v1.TransparencyLogEntry tlog_entries = 3;
+        repeated dev.sigstore.rekor.v1.TransparencyLogEntry tlog_entries = 3 [(google.api.field_behavior) = REQUIRED];
```
haydentherapper, line 58: There’s no guarantee an entry is added to a tlog.

vlsi: Ah, that is interesting. Technically speaking, one can query the log to check the presence, however, it would be nice if the inclusion proof was attached in the first place.

I wonder if it makes sense to add sigstore-specific annotations like 'required for offline bundle'

WDYT?

haydentherapper: There’s still the case where no tlog is used if you’re using only a signed time stamp (in cases of private Sigstore instances). I think we would end up needing multiple annotations, for example how to verity with a timestamp, how to verify with a tlog, how to verify with both. That may be better to have in comments or a part of a reference implementation of verification logic.

protos/sigstore_bundle.proto
```diff
@@ -43,7 +43,7 @@ message TimestampVerificationData {
         // transparency log, or in conjunction for a stronger trust model.
         // Clients MUST verify the hashed message in the message imprint
         // against the signature in the bundle.
-        repeated dev.sigstore.common.v1.RFC3161SignedTimestamp rfc3161_timestamps = 1;
+        repeated dev.sigstore.common.v1.RFC3161SignedTimestamp rfc3161_timestamps = 1 [(google.api.field_behavior) = REQUIRED];
```
kommendorkapten, line 46: The same reasoning is valid here as for the tlog entry, rfc3161 timestamps are optional.

protos/sigstore_common.proto
```diff
@@ -154,13 +154,13 @@ enum SubjectAlternativeNameType {
...
-        SubjectAlternativeNameType type = 1;
+        SubjectAlternativeNameType type = 1 [(google.api.field_behavior) = REQUIRED];
         oneof identity {
                 // A regular expression describing the expected value for
                 // the SAN.
-                string regexp = 2;
+                string regexp = 2 [(google.api.field_behavior) = REQUIRED];
                 // The exact value to match against.
-                string value = 3;
+                string value = 3 [(google.api.field_behavior) = REQUIRED];
```
kommendorkapten, line 163: I have never used `REQUIRED` with a oneof, but I assume this would force a a situation where exactly one field is always set?

kommendorkapten:
Nice, thanks for adding these!

protos/sigstore_bundle.proto
vlsi: If somebody wants to omit rfc3161 timestamp, then they should completely omit `TimestampVerificationData`.
On the other hand, if they add `TimestampVerificationData`, then, I belive, they must provide at least one `rfc3161_timestamps` entry.

What is the purpose of adding `TimestampVerificationData` message with empty contents?

protos/sigstore_common.proto
vlsi: As far as I understand, there's no way to add "required" on a `oneof` itself, so I guess, the convention is "all the oneof fields should be marked with required" which would have a meaning like "`oneof` should have value".

Of course, the client won't be able to fill several fields in `oneof` at the same time.

WDYT?

vlsi:
Can somebody clarify what is the point of storing and comparing the generated code under source control?

protos/sigstore_bundle.proto
kommendorkapten: Sorry, my bad. I read it as the `REQUIRED` was on the actual `TimestampVerificationData`. Ignore this comment :)

protos/sigstore_common.proto
vlsi: Just in case, there is already "all fields are required in `oneof`" in the proto: https://github.com/sigstore/protobuf-specs/blob/16541696de137c6281d66d075a4924d9bbd181ff/protos/sigstore_bundle.proto#L73-L81

kommendorkapten:
> Can somebody clarify what is the point of storing and comparing the generated code under source control?

At least for Go, this is the official way to distribute the Go package, as there are no built packages in Go, the dependency is against a source code repository.

For the other languages, it's to make sure that the code generates properly I would say, to catch an error where the proto files are erroneous.

protos/sigstore_common.proto
kommendorkapten: Nice!

vlsi:
>At least for Go, this is the official way to distribute the Go package, as there are no built packages in Go, the dependency is against a source code repository.

Can it be generated into a `release` or whatever branch?

Frankly speaking, I am not in a position to review and/or propose changes on my behalf for **binary** changes like  https://github.com/sigstore/protobuf-specs/actions/runs/3981004986/jobs/6831243603#step:5:459

kommendorkapten:
I assume you have run `make` locally prior to pushing?

vlsi:
Frankly speaking, I have not.

znewman01:
> Can somebody clarify what is the point of storing and comparing the generated code under source control?

I'm in agreement with you that this is really annoying, sorry about that.

> At least for Go, this is the official way to distribute the Go package, as there are no built packages in Go, the dependency is against a source code repository.
>
> For the other languages, it's to make sure that the code generates properly I would say, to catch an error where the proto files are erroneous.

TBH this is mostly for Go, because Go is extremely opinionated about the release process. We *could* push to a release branch but that's out-of-sync with how everything else gets released.

I don't think there's a really good reason for the other languages to do it; if the goal was to make sure `protoc` is successful, we can just do that in a workflow.

> Frankly speaking, I am not in a position to review and/or propose changes on my behalf for **binary** changes like https://github.com/sigstore/protobuf-specs/actions/runs/3981004986/jobs/6831243603#step:5:459

We have workflows to ensure that they remain in sync, so you can safely ignore these. 

>> I assume you have run `make` locally prior to pushing?
> Frankly speaking, I have not.

Provided you have `docker` installed, all you should need to do for an update is to run `make`.

vlsi:
```
% make all
Building development docker image
docker build -t protobuf-specs-build .
[+] Building 0.1s (9/9) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                                                                                                      0.0s
 => => transferring dockerfile: 37B                                                                                                                                                                                                       0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                         0.0s
 => => transferring context: 2B                                                                                                                                                                                                           0.0s
 => [internal] load metadata for docker.io/namely/protoc-all@sha256:07f1ba9dbe11f5675e2efc8617c9552217dc4c3eb5ccd108f7c3889878dbae50                                                                                                      0.0s
 => [internal] load build context                                                                                                                                                                                                         0.0s
 => => transferring context: 41B                                                                                                                                                                                                          0.0s
 => [1/4] FROM docker.io/namely/protoc-all@sha256:07f1ba9dbe11f5675e2efc8617c9552217dc4c3eb5ccd108f7c3889878dbae50                                                                                                                        0.0s
 => CACHED [2/4] RUN set -ex &&     apt-get update &&     apt-get install -y --no-install-recommends     python3-pip                                                                                                                      0.0s
 => CACHED [3/4] COPY ./dev-requirements.txt /tmp/                                                                                                                                                                                        0.0s
 => CACHED [4/4] RUN python3 -m pip install --upgrade pip &&     python3 -m pip install --requirement /tmp/dev-requirements.txt                                                                                                           0.0s
 => exporting to image                                                                                                                                                                                                                    0.0s
 => => exporting layers                                                                                                                                                                                                                   0.0s
 => => writing image sha256:5a2f6273d08f8049e3c2c1f49c12bfed460cf2a1cb4f1eb9ad693e338c08dc98                                                                                                                                              0.0s
 => => naming to docker.io/library/protobuf-specs-build                                                                                                                                                                                   0.0s
Generating go protobuf files
docker run --platform linux/amd64 -v /Users/.../protobuf-specs:/defs protobuf-specs-build -d protos -l go --go-module-prefix github.com/sigstore/protobuf-specs/gen/pb-go
/usr/bin/find: 'protos': No such file or directory
make: *** [go] Error 1
```

The same goes for just `make`.

My wild guess is it might be caused by the fact I use Apple's M1.

If I change the platform to `linux/arm64/v8`, the error message is different:

```
Generating go protobuf files
docker run --platform linux/arm64/v8 -v /Users/vlsi/Documents/code/protobuf-specs:/defs protobuf-specs-build -d protos -l go --go-module-prefix github.com/sigstore/protobuf-specs/gen/pb-go
Unable to find image 'protobuf-specs-build:latest' locally
docker: Error response from daemon: pull access denied for protobuf-specs-build, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
make: *** [go] Error 125
```

kommendorkapten:
@vlsi That is interesting, I have never seen any issues with that. I'm on macOS/Arm and haven't seen any issue. I'll reach out to you on sigstore slack.

znewman01:
Weird. I am able to do this on an m1 Mac, running macOS with no issues.

> /`usr/bin/find: 'protos': No such file or directory`

This suggests to me that the `protos` directory isn't getting mounted into your running container environment. That's what `-v /Users/.../protobuf-specs:/defs` is supposed to do.

If you run:

```
docker run -it --platform linux/amd64 -v $PWD:/defs --entrypoint /bin/bash protobuf-specs-build 
```

It should drop you in a shell. Do you see your `protobuf-specs` dir mounted if you type `ls`? Specifically, you should see a `protos` directory.

----------------------------------------------------------

If you'd rather not debug this for now, I'm happy to push the generated code for you—just let me know. We can open an issue to discuss changing this workflow as well.

protos/sigstore_rekor.proto
```diff
@@ -88,21 +89,21 @@ message InclusionPromise {
...
-        int64 integrated_time = 4;
+        int64 integrated_time = 4 [(google.api.field_behavior) = REQUIRED];
         // The inclusion promise/signed entry timestamp from the log.
-        InclusionPromise inclusion_promise = 5;
+        InclusionPromise inclusion_promise = 5 [(google.api.field_behavior) = REQUIRED];
         // The inclusion proof can be used for online verification that the
         // entry was appended to the log, and that the log has not been
         // altered.
-        InclusionProof inclusion_proof = 6;
+        InclusionProof inclusion_proof = 6 [(google.api.field_behavior) = REQUIRED];
```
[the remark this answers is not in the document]
vlsi: @kommendorkapten , I'm inclined that the commend should read "The inclusion proof can be used for **offline** verification" rather than "online verification". WDYT?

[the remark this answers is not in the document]
kommendorkapten: The inclusion proof requires that the client performs a request against the transparency log, so it's an online operation. The inclusion _promise_ (i.e verifying the SET) is what can be done offline.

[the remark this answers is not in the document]
vlsi: What is the purpose of having `InclusionProof inclusion_proof` then?
I thought the values in `InclusionProof inclusion_proof` enable `offline` verification.

protos/sigstore_trustroot.proto
```diff
@@ -80,7 +81,7 @@ message TrustedRoot {
         // intermediate and/or leaf certificates.
         // The certificates are intended to be used for verifying artifact
         // signatures.
-        repeated CertificateAuthority certificate_authorities = 3;
+        repeated CertificateAuthority certificate_authorities = 3 [(google.api.field_behavior) = REQUIRED];
```
kommendorkapten, line 84: This one should be optional too. (I think I mentioned that in the previous comment.)

vlsi: Well, if `CertificateAuthority` is optional, then, I guess, the bundle can't have signatures. What do you put into `Bundle.content` if the deployment misses CA?

kommendorkapten: You can, if the artifact is signed by a private key-pair, then there may be no CA involved.

kommendorkapten (dismissed):
Nice!

vlsi:
I'm afraid one more re-generation is needed.

kommendorkapten (dismissed):
Nice

protos/sigstore_rekor.proto
```diff
@@ -40,28 +41,28 @@ message KindVersion {
...
         // the proof was generated.
-        bytes root_hash = 2;
+        bytes root_hash = 2 [(google.api.field_behavior) = REQUIRED];
         // The size of the merkle tree at the time the proof was generated.
-        int64 tree_size = 3;
+        int64 tree_size = 3 [(google.api.field_behavior) = REQUIRED];
         // A list of hashes required to compute the inclusion proof, sorted
         // in order from leaf to root.
         // Not that leaf and root hashes are not included.
         // The root has is available separately in this message, and the
```
haydentherapper, line 60:
```suggestion
        // The root hash is available separately in this message, and the
```

protos/sigstore_rekor.proto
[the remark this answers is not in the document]
haydentherapper: I would make both optional. At least one should be required for this message, but there's no way to easily represent this.

Currently Rekor returns both an inclusion proof and promise always, but I think this behavior shouldn't be reflected in the specs. Future iterations of the log may choose to return either a promise or an inclusion proof or both, but I don't think we should guarantee that a promise is always set. 

Something else to consider is that the log index may not be set for an inclusion promise, because the entry has yet to be admitted into the log. Now, once again the current behavior of Rekor is to always put an entry into the log immediately, but this may not always be the case. I'd rather have clients handle it as optional for now, it'll be less burden in the future if we do make it optional. Thoughts?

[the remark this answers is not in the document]
vlsi: Does that mean that if Rekor omits the proof, then offline verification would not be possible?

By the way, is there something like 'caching Rekor proxy' like there are proxy servers for artifact repositories?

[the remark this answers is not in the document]
haydentherapper: So what's tricky here is there is what Rekor does currently, and what I want Rekor to do in the future. So my first question is, should the protos reflect the current state of the service, or the ideal state?

Slight aside, right now a promise is simply a signed statement from Rekor. Even with a set of hashes to calculate an inclusion proof, there is no guarantee that an entry actually exists in Rekor without doing an online lookup. This is because you need to verify consistency (that the log is still append-only), which would require an online lookup. I've been thinking about how to improve this, but we're awhile away from being able to have stronger offline verification.

With what we've got right now, a client could "verify" offline using either the hashes from the proof or the promise, they're pretty much equivalent. Currently Rekor will return both because of its design, but technically it could omit one or the other - A promise would be omitted when a proof is requested (because if a proof can be generated, why need a promise?), or a proof would be omitted when only a promise can be generated if the entry is actively being uploaded (Rekor operates in a batched mode, but the batch size = 1 currently, so entries are immediately uploaded)

So I don't really know the best way to handle this for whether it should be required or optional. I'd lean towards being permissive and allow either the proof or promise or both, but not require that a promise is always present.

For proxies, we've chatted a little bit about Rekor read-only replicas. I see this as a requirement for being able to scale out Rekor. This is definitely something I want to do, we just need to scope it out and figure out technical blockers.

protos/sigstore_bundle.proto
```diff
@@ -63,7 +63,7 @@ message VerificationMaterial {
 message Bundle {
         // MUST be application/vnd.dev.sigstore.bundle+json;version=0.1
         // when encoded as JSON.
-        string media_type = 1;
+        string media_type = 1 [(google.api.field_behavior) = REQUIRED];
```
woodruffw, line 66: Possibly irrelevant, but is anybody familiar with what this annotation implies about non-JSON-serialized protobufs? My understanding is that `media_type` is only sensible when encoded as JSON, so it may not make sense to set it as `REQUIRED` here.

woodruffw (commented):
LGTM mod one comment

protos/sigstore_rekor.proto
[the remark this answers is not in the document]
kommendorkapten: @haydentherapper I think this is good as-is now, as it reflects the current way Rekor works? 

As there are more changes expected in Rekor (SET v2), omitting the log index I think we can wait with them until Rekor is updated?

protos/sigstore_bundle.proto
kommendorkapten: Let me try it out!

kommendorkapten: So I was digging around a bit, and was not able to get a program to fail marshalling to either binary or JSON when the preconditions where not met (i.e some `REQUIRED` fields was not present). This led me to the [documentation](https://google.aip.dev/203) where I found this:

> **Note**: The vocabulary given in this document is for descriptive purposes only, and does not itself add any validation. The purpose is to consistently document this behavior for users.

So it any developer would still need to perform any validation manually it seems. The generated (Go) code does not seem to include any annotations for JSON marshal/unmarshal that would imply that a given field is `REQUIRED` or not.

Maybe this is old news to everyone, but it was not for me (my previous example is to rely on `optional` keywords and treat anything else as `REQUIRED`).

But to go back to @woodruffw 's question. I think it still make sense to include the `media_type` even when binary marshalling is being conducted, as we may not know what a downstream consumer is intending to do with the message. Maybe there is a service pulling the data from a central registry and serializing to disk (via JSON) for further consumption by a different service?

kommendorkapten: @haydentherapper you got any different experiences with `REQUIRED` fields?

haydentherapper: This is correct, we need to find a library for enforcing this (I know of the internal google one, not sure what’s available for OSS)

vlsi: Java client does not verify `REQUIRED` fields, however, there's API to get the annotations: https://github.com/sigstore/sigstore-java/pull/317/files#diff-4a847f4c6f9e37621c0d9d5da4a86c027037a2c63052205b6471a662724e5b73R130-R136

haydentherapper: https://github.com/mwitkow/go-proto-validators - this looks a lot like the internal library

asraa: I used protoc-gen-validate in an old life.

znewman01 (commented):
My vote is to err on the side of not marking things required for now.

But I suppose an error either way doesn't really hurt here, since it's not really enforced in proto and these things being missing should cause other verification errors anyway.

protos/sigstore_bundle.proto
asraa: What if it were a fixed string? Or at least an enum? That maybe prevents clients or clarifies usage in code that it must be certain values. 

At the same time, I do want it as a type hint. So yeah, maybe not required, but otherwise I don't know how to package this unless it's inside another wrapped proto with (payload=bundle, media_type)

woodruffw: Agreed -- IMO it's not ideal, but the alternatives (hooking into the marshalling, or doing it after marshalling) are also probably more error prone/more likely to be forgotten due to a lack of controls.

I'm loathe to add *yet another* layer of schematization here, but maybe what we really need is a JSON Schema definition that gets generated from these protobuf definitions. That schema could then include (and enforce) the requirement hints. 

Just from a brief look, it seems like https://github.com/chrusty/protoc-gen-jsonschema might do the trick, and has support for their own `required` annotation to mark fields as non-optional.

woodruffw: Works for me. This is definitely not something that should block any other work here, and I can pursue it as an independent follow-up.

kommendorkapten:
> But I suppose an error either way doesn't really hurt here, since it's not really enforced in proto and these things being missing should cause other verification errors anyway.

This is good I think. As we have `REQUIRED` on some fields already, I think adding more makes it a bit more consistent and clear for a user. When we go after #67 and have a strong test harness we can nail it down to perfection.

I think the state is now when some fields are marked as required and some now (which are) is a bit confusing. So either we make a "best effort" like this PR, or remove them all together until we can offer a good way to programatically verify a message.

woodruffw:
I'm +1 for getting this merged -- I agree that these annotations have limited value on the programmatic side, but they _do_ have some value on the design/engineering side (in terms of visually confirming my intuition that various fields actually are required).

In particular, it's nice to have a confirmation that the SET and log entry body are required, while the inclusion proof (intuitively) isn't!

CC @tnytown

kommendorkapten (approved):
I'm ok too with merging this.
We **don't** need to cut a breaking release, we can do a patch update as this is only clarifications and does not change the original intent.

edit: sentence missed a crucial word, *don't*, thanks @feelepxyz for pointing that out 🚀

feelepxyz:
> We need to cut a new breaking release

@kommendorkapten to clarify, you mean we _don't_ need to do a major version/breaking change release?

woodruffw:
#79 partially overlapped with this, so we'll need to rebase/deconflict here 🙂

haydentherapper:
Closing since we've made a lot of fields required already.

Issue #72, opened 2023-03-23
FR: Separate canonicalized_body from TransparencyLogEntry?


wlynch:
@asraa @haydentherapper and I had a discussion over at https://github.com/sigstore/rekor/pull/1390#discussion_r1139350482 that I figured is worth opening up an issue here for discussion.

I'm currently implementing offline verification for Gitsign. Currently this is done by serializing all the fields in TransparencyLogEntry individually to their own OIDs, then reconstructing the message at verification time. An idea was suggested to encode the serialized version of TransparencyLogEntry instead, but the message as defined today is problematic because of `canonicalized_body`.

`canonicalized_body` was added to TransparencyLogEntry in https://github.com/sigstore/protobuf-specs/pull/10 to let clients verify data without needing to reconstruct the body themselves due to concerns around normalization of DSSE encoding (https://github.com/sigstore/protobuf-specs/issues/9).

However, for Gitsign we would like to omit this body and reconstruct it.
Gitsign has constraints on the signature format to stay in-line with other Git signatures. If we store a serialized version of the TransparencyLogEntry with canonicalized_body, we end up doubling the size of the signature for data that we already store elsewhere and can reconstruct.

I'd like to propose modifying TransparencyLogEntry to allow only storing the minimal data needed. Few options I can see:

1. We break out canonicalized_body from TransparencyLogEntry and create a new wrapper message that combines them. e.g.

    ```proto
    // name tbd - open to bikeshedding
    message TransparencyLogEntryBundle {
      TransparencyLogEntry entry = 1;
      bytes canonicalized_body = 2;
    }
    ```

    This way the TLE + canonicalized_body can still be included in the full bundle, but applications that only need the minimal data in TransparencyLogEntry can just use that.

2. We add as part of the spec that it's valid for canonicalized_body field to be unset, and if it is it's the responsibility of the client to construct it during verification. This unfortunately has the side effect of effectively undoing the changes in #10, since clients will need to be able handle cases where the canonicalized_body is not present.

3. We let Gitsign be special and document our off-label usage of TransparencyLogEntry there. I'm a little hesitant about this since a) we'd wouldn't be able to treat TransparencyLogEntry like an opaque envelope and may need to track changes overtime and b) if gitsign is running into this my guess is other signing tools may as well.

Also open to other ideas!

cc @znewman01 @codysoyland

znewman01:
I like option 1. It's breaking-ish so we'd need a version bump.

[...]

PR #74, opened 2023-03-30
Clarify usage of VerificationLogEntry.canonicalized_body.

wlynch:
#### Summary

This updates the doc comment of canonicalized_body to make it clear that it is an optional field, and provides more details on when the field should be set / when it can be safely omitted.

This is not intended to be a change in behavior - this is just clarifying existing intent of the field. 

Fixes https://github.com/sigstore/protobuf-specs/issues/72

#### Release Note

- Updated the description of VerificationLogEntry.canonicalized_body to clarify when it should be set / when it can be safely omitted.

#### Documentation

If there's anything else I need to do besides `make all`, let me know!

### DIFF
--- a/protos/sigstore_rekor.proto
+++ b/protos/sigstore_rekor.proto
@@ -103,8 +103,8 @@ message TransparencyLogEntry {
         // entry was appended to the log, and that the log has not been
         // altered.
         InclusionProof inclusion_proof = 6;
-        // The canonicalized transparency log entry, used to reconstruct
-        // the Signed Entry Timestamp (SET) during verification.
+        // Optional. The canonicalized transparency log entry, used to
+        // reconstruct the Signed Entry Timestamp (SET) during verification.
         // The contents of this field are the same as the `body` field in
         // a Rekor response, meaning that it does **not** include the "full"
         // canonicalized form (of log index, ID, etc.) which are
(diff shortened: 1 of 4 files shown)


protos/sigstore_rekor.proto
```diff
@@ -103,18 +103,23 @@ message TransparencyLogEntry {
...
+        // produced determinisitically (e.g. inconsistent JSON field ordering,
+        // differing whitespace, etc).
         //
-        // Clients MUST verify that the signatured referenced in the
+        // If set, clients MUST verify that the signature referenced in the
         // `canonicalized_body` matches the signature provided in the
         // `Bundle.content`.
+        // If not set, clients are responsible for constructing an equivalent
+        // payload from other sources to verify the signature.
         bytes canonicalized_body = 7;
```
codysoyland, line 124: Should this have the `optional` keyword?

```suggestion
        optional bytes canonicalized_body = 7;
```

wlynch: ¯\_(ツ)_/¯ I'm in different.

proto3 generally takes a stance of everything is "optional", i.e. it's not considered an error if you don't set a field, it just won't include it in serialization if it's the zero value. IIUC the optional keyword for proto3 is really for field existence if you want to distinguish between the field wasn't set vs the empty string was set, which I don't think gives us much value in this context.

znewman01 (approved):

codysoyland (approved):

haydentherapper (approved):

PR #76, opened 2023-04-24
gen, protos: tweak envelope docs

woodruffw:
This just renames `body` in the DSSE envelope docs to reflect the actual field name, which is `payload`.

CC @znewman01 @kommendorkapten


znewman01 (approved):

kommendorkapten:
Nice. You man want to create a similar PR against the upstream repo where the file is sourced from: https://github.com/secure-systems-lab/dsse/blob/master/envelope.proto

woodruffw:
Thanks for pointing that out! Opened https://github.com/secure-systems-lab/dsse/pull/56.

PR #77, opened 2023-04-24
gen, protos: document chain membership policy

### LINKED ISSUE #75: Removing the root cert from the bundle's cert chain?
I was thinking about this some more:

https://github.com/sigstore/protobuf-specs/blob/4dbf10bc287d76f1bfa68c05a78f3f5add5f56fe/protos/sigstore_common.proto#L167-L175

...and I think, as specified, it presents a sharp edge towards Sigstore clients: the naive thing to do would be to shove the entire cert chain into the verifying context (i.e. a thing that looks roughly like an OpenSSL `X509_STORE`). However, if a client does this, then Sigstore signatures effectively become self-signed: any malicious user can insert a cert chain that includes a root CA that isn't one of the trusted ones, and chain validation will resolve against that CA instead.

Clients still need to handle this their own (they should verify the expected contents of the bundle's cert chain), but as a risk reduction strategy: what does everyone think about specifying the root certificate out of the bundle's cert chain? That would turn this from a silent "gotcha" into a spec violation, one that we could write conformance tests against.

woodruffw:
Closes #75.

CC @znewman0, @haydentherapper, @kommendorkapten

### DIFF
--- a/protos/sigstore_common.proto
+++ b/protos/sigstore_common.proto
@@ -169,9 +169,16 @@ message SubjectAlternativeName {
 message X509CertificateChain {
         // The chain of certificates, with indices 0 to n.
         // The first certificate in the array must be the leaf
-        // certificate used for signing. Any intermediate certificates
-        // must be stored as offset 1 to n-1, and the root certificate at
-        // position n.
+        // certificate used for signing.
+        //
+        // Signers MUST NOT include their root CA or CAs in their embedded
+        // certificate chains, and SHOULD NOT include intermediate CAs that
+        // appear in independent roots of trust.
+        //
+        // Verifiers MUST validate the chain carefully to ensure that it
+        // upholds the signer's invariants, and MUST be prepared to handle
+        // old or non-complying bundles by ignoring all root CAs and any
+        // intermediate CAs that appear in independent roots of trust.
         repeated X509Certificate certificates = 1;
 }
 
--- a/gen/pb-go/common/v1/sigstore_common.pb.go
+++ b/gen/pb-go/common/v1/sigstore_common.pb.go
@@ -874,9 +874,16 @@ type X509CertificateChain struct {
 
 	// The chain of certificates, with indices 0 to n.
 	// The first certificate in the array must be the leaf
-	// certificate used for signing. Any intermediate certificates
-	// must be stored as offset 1 to n-1, and the root certificate at
-	// position n.
+	// certificate used for signing.
+	//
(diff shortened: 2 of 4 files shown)


protos/sigstore_common.proto
```diff
@@ -169,9 +169,16 @@ message SubjectAlternativeName {
...
-        // must be stored as offset 1 to n-1, and the root certificate at
-        // position n.
+        // certificate used for signing.
+        //
+        // Signers MUST NOT include their root CA or CAs in their embedded
+        // certificate chains, and SHOULD NOT include intermediate CAs that
+        // appear in independent roots of trust.
+        //
+        // Verifiers MUST validate the chain carefully to ensure that it
+        // upholds the signer's invariants, and MUST be prepared to handle
```
znewman01, line 179: "upholds the signer's invariants" is a little unclear to me. Also, one person's root CA is another person's intermediate CA, so I'd stress "signer" a little bit less.

Maybe:

> Verifiers MUST validate the chain carefully to ensure that it chains up to a root CA that they trust, regardless of whether the chain includes additional intermediate/root CAs. Verifiers MAY enforce additional constraints, such as requiring that all intermediate CAs appear in an independent root of of trust.
>
> Verifiers SHOULD handle old or non-complying bundles that have additional intermediate/root CAs.

znewman01 (changes requested):

protos/sigstore_common.proto
```diff
@@ -169,9 +169,16 @@ message SubjectAlternativeName {
 message X509CertificateChain {
         // The chain of certificates, with indices 0 to n.
         // The first certificate in the array must be the leaf
-        // certificate used for signing. Any intermediate certificates
-        // must be stored as offset 1 to n-1, and the root certificate at
-        // position n.
+        // certificate used for signing.
+        //
+        // Signers MUST NOT include their root CA or CAs in their embedded
```
haydentherapper, line 174: nit, CA -> CA certificates here and below (let's just use the plural)

protos/sigstore_common.proto
woodruffw: Sounds good!

protos/sigstore_common.proto
woodruffw: Done!

protos/sigstore_common.proto
woodruffw: Done!

znewman01 (approved):

kommendorkapten:
Nice one!

PR #78, opened 2023-04-24
Clarify TimeRange validity periods

znewman01:


loosebazooka (approved):
nice, thanks!

haydentherapper (approved):

kommendorkapten (changes requested):
This comments are now in conflict with the definition of a `TimeRange`:

https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L190

> // The time range is half-open and does not include the end timestamp,
> // i.e [start, end).

We should pick one definition and be consistent IMO.

I would be in favour of using the already defined (Half open), but are open for discussion.

edit: fixed a broken sentence.

kommendorkapten:
sigstore-js is *inclusive* for both certificate and transparency log verification:
https://github.com/sigstore/sigstore-js/blob/b4e96a2df70abbcb0cc210f50568de04d46c23e8/src/ca/verify/chain.ts#L74
https://github.com/sigstore/sigstore-js/blob/b4e96a2df70abbcb0cc210f50568de04d46c23e8/src/tlog/verify/set.ts#L114

So this may indicate that we should align with RFC 5280.

@woodruffw any details on how Python client?

znewman01:
Good catch. My vote is to align with RFC 5280. I've updated the `TimeRange` comment.

Fortunately, it doesn't *actually* matter—a nanosecond should never make the difference in normal use.

haydentherapper (approved):

protos/sigstore_trustroot.proto
```diff
@@ -52,7 +52,10 @@ message CertificateAuthority {
         dev.sigstore.common.v1.X509CertificateChain cert_chain = 3;
         // The time the *entire* chain was valid. This is at max the
         // longest interval when *all* certificates in the chain were valid,
-        // but it MAY be shorter.
+	// but it MAY be shorter. Clients MUST check timestamps against *both*
+	// the `valid_for` time range *and* the entire certificate chain.
```
woodruffw, line 56: Weird spacing/indentation here? 🙂

woodruffw:
> @woodruffw any details on how Python client?

Looks like we actually have a small bug with this: we're inclusive on `notAfter` but not on `notBefore`. I'll fix that now.

woodruffw:
> Good catch. My vote is to align with RFC 5280. I've updated the `TimeRange` comment.

Strong agree!

protos/sigstore_trustroot.proto
znewman01: whoops, was tabs

haydentherapper (approved):

kommendorkapten (approved):
Nice!

PR #83, opened 2023-06-30
gens, protos: initialize rust codegen

jleightcap:
#### Summary

Towards Bundle support in sigstore-rs, per https://github.com/sigstore/sigstore-rs/issues/280#issuecomment-1609979744.

#### Release Note

Adds rust codegen.


gen/pb-rust/build.rs
jleightcap, line 1: I went with the approach of using a build script. However, I'm not super familiar with the packaging of library crates -- need to verify that:

- protoc dependency doesn't leak,
- these relatives paths all work.

Makefile
```diff
@@ -49,6 +49,16 @@ ruby: docker-image
...
 		-c "cd ./gen/pb-ruby && protoc -I/opt/include -I../../protos/ --ruby_out=lib ../../protos/*.proto"
 
+rust: docker-image
+	@echo "Generating rust protobuf files"
+	docker run \
+		--platform linux/amd64 \
+		-v ${PWD}:/defs \
+		-e "RUST_BACKTRACE=1" \
+		--entrypoint bash ${PROTOC_IMAGE} \
+		-c "cd gen/pb-rust && cargo build"
```
jleightcap, line 59: This isn't the best. Because some of the includes are in `/opt/`, any cargo commands have to be threaded through this (or parallel) Makefile rules.

gen/pb-rust/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+pub mod envelope;
```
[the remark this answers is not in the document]
woodruffw: Making sure I understand: this file was manually written, right? 

In that case, yeah -- ideally the `lib.rs` would also be generated by our codegen; we don't want to accidentally forget to expose new protobuf modules as they're added.

gen/pb-rust/codegen/Cargo.toml
jleightcap, line 1: This is the most KISS method I could come up with for AOT codegen. This has a downside of decoupling the build steps -- the codegen uses the top-level `Dockerfile`, but packaging uses `gen/pb-rust/Cargo.toml`. However, a `make all` for any new `.proto`s should cover this case.

woodruffw:
Making sure I understand how this works:

1. The separate `codegen` crate performs the AOT codegen
2. The AOT codegen is done *into* the `sigstore_protobuf_specs`, which is the crate that users are expected to actually use

Am I getting that right? If so, that seems reasonable and this LGTM; CI needs fixing 🙂

woodruffw:
Suggestion: squash down to a single commit and do a sign-off from there, to get DCO passing. You could also rebase and sign-off for each commit separately, but that seems painful given that we'll probably just squash it anyways 🙂

jleightcap:
> Making sure I understand how this works:
> 
> 1. The separate `codegen` crate performs the AOT codegen

Yep, 

> 2. The AOT codegen is done _into_ the `sigstore_protobuf_specs`, which is the crate that users are expected to actually use

and the `codegen` crate populates the parent pb-rust crate.

woodruffw (approved):
LGTM, nice work!

woodruffw:
CC @kommendorkapten and @haydentherapper for review/approval as well, since I can't merge.

haydentherapper:
Thanks! Can you also update https://github.com/sigstore/protobuf-specs/blob/main/RELEASE.md?

Dockerfile
```diff
@@ -6,6 +6,14 @@ RUN set -ex && \
     apt-get install -y --no-install-recommends \
     python3-pip
 
+# Install Rust cargo.
+RUN set -ex && \
+    apt-get install -y --no-install-recommends \
+        curl \
+        build-essential
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- -y
```
kommendorkapten, line 14: Isn't there a better way to install the rust toolchain than `curl | bash` as root?

Makefile
```diff
@@ -49,6 +49,16 @@ ruby: docker-image
 		--entrypoint bash ${PROTOC_IMAGE} \
 		-c "cd ./gen/pb-ruby && protoc -I/opt/include -I../../protos/ --ruby_out=lib ../../protos/*.proto"
 
+rust: docker-image
+	@echo "Generating rust protobuf files"
+	docker run \
+		--platform linux/amd64 \
```
kommendorkapten, line 55: I'm not super familiar with Rust yet, but I assume this would only create a package for `linux/amd64`? Would we want to support other OS/architectures too?

kommendorkapten: Oh, reading more it seems this is just the target OS/arch for the codegen, so when the docker image has run successfully, only Rust source code is being generated, no object files?

Dockerfile
woodruffw: We could probably install the Debian/Ubuntu distribution instead, but IMO this will serve us better over time: the system distribution tends to lag behind the latest stable, and the Rust dependency ecosystem (unfortunately?) tends to adopt stabilized features relatively quickly.

I think the risk of doing so is similar to the risk of installing a bunch of Python development dependencies as root, which we do immediately below as well 🙂

Makefile
woodruffw: > Oh, reading more it seems this is just the target OS/arch for the codegen, so when the docker image has run successfully, only Rust source code is being generated, no object files?

Correct! I think that `--platform` option actually shouldn't be needed at all; @jleightcap may have added it because of his local development environment.

(Rust package distributions are always source only, so the environment that creates them should have no effect on their buildability across other platforms.)

Dockerfile
jleightcap: For better or worse this is the default recommended installation per https://www.rust-lang.org/tools/install. `cargo` is available, but not `rustup` from `apt`. However, the `cargo` version is out of date:


```
this version of Cargo is older than the `2021` edition, and only supports `2015` and `2018` editions.
```

And the only method of updating `cargo` AFAIK is via `rustup`.

Makefile
jleightcap: Ah, this is a copy-paste from the other targets ([python](https://github.com/sigstore/protobuf-specs/blob/ae23636772323882df1dabd6d4cd42793ed211bb/Makefile#L30), [typescript](https://github.com/sigstore/protobuf-specs/blob/ae23636772323882df1dabd6d4cd42793ed211bb/Makefile#L39), &c).

The [`pb-rust` crate itself](https://github.com/sigstore/protobuf-specs/blob/ae23636772323882df1dabd6d4cd42793ed211bb/gen/pb-rust/Cargo.toml) is architecture agnostic.

woodruffw (approved):

woodruffw:
CC @kommendorkapten and @haydentherapper for re-review 🙂

haydentherapper (approved):
@kommendorkapten for approval and merge

kommendorkapten (approved):

woodruffw:
Thanks all! Great work here @jleightcap.

haydentherapper:
+1, thank you for your contribution!

PR #86, opened 2023-07-08
Include CloudEvents proto to generated bundle

jalseth:
#### Summary

Sigstore will follow the CloudEvents specification when publishing events to Pub/Sub topics. Protobuf events can be used for better efficiency compared to JSON.

#### Release Note

* Added CloudEvents message to be used by Sigstore projects when publishing events.

#### Documentation

N/A, all docs on events usage will be included in the relevant project (rekor, rekor-monitor, etc.).


protos/events.proto
```diff
@@ -0,0 +1,81 @@
...
+ * CloudEvent Protobuf Format
+ *
+ * - Required context attributes are explicity represented.
+ * - Optional and Extension context attributes are carried in a map structure.
+ * - Data may be represented as binary, text, or protobuf messages.
+ */
+
+ syntax = "proto3";
+
+ package io.cloudevents.v1;
```
bobcallaway, line 25: should this be `dev.sigstore.rekor.events.v1` or a `dev.sigstore.events` package name?

woodruffw: > This is the exact proto definition from https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/cloudevents.proto with no changes.

Sorry if I'm missing something, but: if this is an exact copy of another specification and isn't specific to Sigstore, then what's the advantage of having it in this repository/generated in these packages? Is the idea that it'll eventually be referenced in other Sigstore-specific messages?

The DSSE envelope is a bit of a strange case: it's copied from another repository but gets used directly in the `Bundle` definition. IMO it really should be under the `dev.sigstore` namespace as well, but we can't change that without another breaking release 🙂

bobcallaway: lines 30 & 31 appear to be customized though to sigstore... so it feels wrong to have that in someone else's namespace?

jalseth: @woodruffw The CloudEvents repo does not provide generated code that can be referenced, so including it here seemed like a good location for use among Sigstore projects. 

The initial use case is https://github.com/sigstore/rekor/pull/1580 for publishing events about new entries in Rekor. If future use cases warrant it, Rekor or Fulcio may publish other events and can reuse this.

woodruffw: Makes sense, thanks for the explanation @jalseth!

I _think_ I agree with @bobcallaway here: IMO any amount of tailoring for Sigstore means that this should live under the `dev.sigstore` namespace, even if doing so effectively means some duplication/blurred lines around what is or isn't a Sigstore message definition.

(I think this also applies to the current in-toto messages, and I'm going to file a separate issue for us to consider re-homing those under `dev.sigstore`.)

jalseth: Sounds good, either is fine with me. The descriptors aren't included in the wire format, so the message name (including the package) is up to the client with whatever descriptors they have loaded when they deserialize it.

bobcallaway (approved):

jalseth:
Not sure why the check failed. I used `make all` to generate the code for each language and this only adds new code, it does not modify any existing code.

haydentherapper:
Have you rebased against main?

jalseth:
Oh, thanks! I didn't realize 20+ commits had been made since I opened this. Should be fixed now.

haydentherapper (approved):

haydentherapper:
Thanks @jalseth!

PR #87, opened 2023-07-10
Use ruby_package proto option

pwelch:
#### Summary

This fixes the Ruby module names generated from the proto files by using the `option ruby_package=`. Setting these allows for more idiomatic Ruby class naming within the namespace `Sigstore::`.

#### Release Note

- There is a breaking change in the Ruby class names. All Ruby class references will be in the `Sigstore::` namespace now. Please reference the generated Ruby files for the new class names.


haydentherapper (approved):
That's strange that it required updating all generated files, but I don't see anything problematic.

pwelch:
> That's strange that it required updating all generated files, but I don't see anything problematic.

@haydentherapper Agreed. I assumed only the Ruby generated code would need to be updated but it looks like it updates something on several of the generated files?

PR #99, opened 2023-07-25
Update protoc, add dependabot

haydentherapper:
I was running into an error with the Docker container updating its apt cache, updating to the latest protoc image fixed it.

Also added dependabot to keep it up to date.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ clean:
 		gen/pb-typescript/src/__generated__ \
 		gen/pb-python/sigstore_protobuf_specs/dev \
 		gen/pb-python/sigstore_protobuf_specs/io \
-		gen/pb-rust/target
+		gen/pb-rust/codegen/target
 	docker rmi -f ${PROTOC_IMAGE}
 
 help:
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,8 @@ clean:
 		gen/pb-typescript/src/__generated__ \
 		gen/pb-python/sigstore_protobuf_specs/dev \
 		gen/pb-python/sigstore_protobuf_specs/io \
-		gen/pb-rust/codegen/target
+		gen/pb-rust/codegen/target \
+		gen/pb-rust/target
 	docker rmi -f ${PROTOC_IMAGE}
 
 help:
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -79,7 +79,7 @@ clean:
 		gen/pb-typescript/src/__generated__ \
 		gen/pb-python/sigstore_protobuf_specs/dev \
 		gen/pb-python/sigstore_protobuf_specs/io \
-		gen/pb-rust/target
+		gen/pb-rust/codegen/target
```
haydentherapper, line 82: @woodruffw Could you confirm this is fine? `target` seemed to be nested under codegen.

haydentherapper:
@woodruffw One other thing I noticed, there's a file permissions issue with the created codegen for rust. When I run make again, I get:

```
Building development docker image
docker build -t protobuf-specs-build .
error checking context: no permission to read from '/usr/local/google/home/hblauzvern/protobuf-specs/gen/pb-rust/codegen/target/debug/incremental/sigstore_protobuf_specs_codegen-219stdjo4fqmo/s-gn63o9g6ki-d2358w.lock'
make: *** [Makefile:66: docker-image] Error 1
```

Do you know where I should update for this?

haydentherapper:
There's reports of this issue in https://github.com/rust-lang/cargo/issues/6757, so I assume we just need to chown some directory

Makefile
woodruffw: I _believe_ that's fine, although we might want both -- `codegen` is its own sub-crate.

CC @jleightcap for confirmation.

woodruffw:
> One other thing I noticed, there's a file permissions issue with the created codegen for rust. When I run make again, I get:

Hmm, I haven't seen this before. I wonder if this is because `cargo` is trying to read/write to a directory that's shared with the host? I'll try and repro tomorrow.

kommendorkapten:
> I was running into an error with the Docker container updating its apt cache, updating to the latest protoc image fixed it.

Did you try running `make docker-image-no-cache` @haydentherapper ?

I was running into similar issues so that's why I added the target drop the caches.

Makefile
haydentherapper: Thanks, added this back.

PR #109, opened 2023-07-26
Let docker clean up generated folder for rust

haydentherapper:
Docker generates files and folders as root when docker is run as root, which is the default. This means that a developer needs to sudo chown the target directory when starting from scratch, or else they can't rerun make rust, which will throw a permissions error.


haydentherapper:
Confirmed that this resolves the issue of running `make` repeatedly.

woodruffw (approved):
Nice, thanks @haydentherapper!

Issue #67, opened 2023-02-08
Generate JSON Schema from Protobuf definitions and use it to enforce fields?


woodruffw:
Breakout from https://github.com/sigstore/protobuf-specs/pull/64#discussion_r1100287104: I think one possible solution to our "required" field woes is as follows:

1. Keep the current Protobuf definitions, and mark fields we expect/need as required;
2. Use a Protobuf to JSON Schema generator to generate a corresponding JSON Schema, including actual field enforcement (meaning that the generator will need to respect/handle the relevant annotations)
3. Use the generated JSON Schema to actually enforce the validity of JSON marshaled messages

This shouldn't be too difficult to do, and it looks like there's already some tooling out there that does it (<https://github.com/chrusty/protoc-gen-jsonschema>). 

OTOH, it adds _yet another_ layer of schematization, and only applies to JSON-marshaled messages. 

cc @znewman01

kommendorkapten:
An alternative approach: the protobuf files are not expected to change that frequently. How about we hand code validate functions for each ecosystem, as part of the language bindings in this repo? 
Wire up conformance testing to the automation. It will be "expensive" upfront, but I think will simplify our lives down the road with much less complexity.

znewman01:
Tbh I think we could get most of the benefits here by just making sure the conformance tests check for a proper media type, then each language can do it however they’d like.(I think all of the prior ideas are “better” but want to make sure we spend time on the most impactful issues :) )On Feb 9, 2023, at 4:15 AM, Fredrik Skogman ***@***.***> wrote:﻿
An alternative approach: the protobuf files are not expected to change that frequently. How about we hand code validate functions for each ecosystem, as part of the language bindings in this repo?
Wire up conformance testing to the automation. It will be "expensive" upfront, but I think will simplify our lives down the road with much less complexity.

PR #112, opened 2023-07-28
Added a prototype for generating jsonschema files

kommendorkapten:
WIP to generate JSON schema files, as coined by @woodruffw (see Slack: https://sigstore.slack.com/archives/C03SZ6SHU90/p1690490585606889)

Closes https://github.com/sigstore/protobuf-specs/issues/67

Consider this PR WIP so far.

There are some general topics I would like to bring up for discussion too related to the current Docker image we are using for building the protoc compiler that I discovered when working on this. I'll create a separate issue for that.

#### Release Note
Added support for generating JSON schema files

#### Documentation
N/A


kommendorkapten:
The reason I added a new Docker image is that namely/protoc-all is too old to build newer Go programs, see issue #113

woodruffw (approved):
This looks pretty good! I did a quick pass over the generated schemas and nothing looks _way_ off 🙂 

I've asked @jleightcap to take a look at using these to generate the Rust crate, once he's finished with some conformance suite work.

Dockerfile.jsonschema
```diff
@@ -0,0 +1,6 @@
+# 3.18.2
+FROM alpine@sha256:25fad2a32ad1f6f510e528448ae1ec69a28ef81916a004d3629874104f8a7f70
+RUN apk add --update protoc protobuf-dev go git
```
haydentherapper, line 3: Should we be pinning versions for these tools?

woodruffw: I think so, yeah -- IIRC we've seen `protoc` make codegen changes between releases before, so pinning here may prevent rude CI discrepancies 🙂

Makefile
codysoyland, line 1: Should we add `jsonschema` to the list of targets for [`make all`](https://github.com/sigstore/protobuf-specs/blob/main/Makefile#L19)? This way, it will become [part of the CI check](https://github.com/sigstore/protobuf-specs/blob/main/.github/workflows/generate.yml) to make sure generated files are always updated.

kommendorkapten:
Yeah, we should pin the versions and make sure we perform this build for
the all target. I’ll be away for vacation now for two weeks, but can take a
stab when I’m back, and looking into creating a single new image that
covers all languages so we can drop the namely image.

If any of you are in a hurry to get the JSON schema in now, feel free to
modify this PR. Maintainers will have permission to do so.

woodruffw:
> If any of you are in a hurry to get the JSON schema in now, feel free to modify this PR. Maintainers will have permission to do so.

I'm going to have @jleightcap work off of the schema in this PR, so I'll take over! Enjoy your vacation!

Dockerfile.jsonschema
```diff
@@ -0,0 +1,8 @@
+# 3.18.2
+FROM alpine@sha256:25fad2a32ad1f6f510e528448ae1ec69a28ef81916a004d3629874104f8a7f70
+RUN apk add --update protoc=3.21.12-r2 protobuf-dev=3.21.12-r2 go=1.20.5-r0 git=2.40.1-r0
+RUN go install github.com/chrusty/protoc-gen-jsonschema/cmd/protoc-gen-jsonschema@1.4.1
+# This is required to get the field_behavior.proto file
+# NOTE: --filter=tree:0 performs a treeless clone; we do this to optimize cloning
+# this otherwise relatively heavy repository.
+RUN git clone --filter=tree:0 https://github.com/googleapis/googleapis.git
```
woodruffw, line 8: NB: This repository doesn't appear to be versioned in any way, and isn't packaged in alpine. I could check out a fixed revision if we're like more consistency here.

Makefile
woodruffw: Done!

Dockerfile.jsonschema
```diff
@@ -0,0 +1,8 @@
+# 3.18.2
+FROM alpine@sha256:25fad2a32ad1f6f510e528448ae1ec69a28ef81916a004d3629874104f8a7f70
+RUN apk add --update protoc=3.21.12-r2 protobuf-dev=3.21.12-r2 go=1.20.5-r0 git=2.40.1-r0
```
woodruffw, line 3: NB: I picked these version numbers not because they're the most recent, but because they're consistent with Alpine's `world` version for the Alpine version chosen. More recent versions _could_ probably be made to work, but I figured this was the path of least resistance.

Dockerfile.jsonschema
znewman01: I'd like to pin it to a fixed revision.

znewman01 (dismissed):
This LGTM, except I would like the googleapis/googleapis repo pinned too :)

Dockerfile.jsonschema
woodruffw: Hmm, this is moderately annoying: we can't do a fast-path `clone` with a specific revision. I'll see about just downloading it directly as a ZIP or similar and unarchiving.

Dockerfile.jsonschema
```diff
@@ -0,0 +1,10 @@
...
+RUN go install github.com/chrusty/protoc-gen-jsonschema/cmd/protoc-gen-jsonschema@1.4.1
```
haydentherapper, line 4: Do you know if dependabot will suggest updates for Go dependencies in Dockerfiles? Otherwise we might want to add a hack like https://github.com/sigstore/rekor/blob/main/hack/tools/go.mod

znewman01 (approved):

haydentherapper:
Should the jsonschema be versioned and have releases cut along with the other languages? Otherwise, it'll always be from HEAD. We could have folders for each version, /schema/v0.2/<files>? Not sure if there's precedent for this.

woodruffw:
> Should the jsonschema be versioned and have releases cut along with the other languages? Otherwise, it'll always be from HEAD. We could have folders for each version, /schema/v0.2/? Not sure if there's precedent for this.

Hmm, good call -- we want to use this as the building block for the Rust bindings, so we should probably version it. IMO that can be done either by tagging (like the other bindings) or by having directories for either version; I don't think it'll make a major difference to the Rust bindings either way.

haydentherapper:
Is it sufficient to use an existing tag like https://github.com/sigstore/protobuf-specs/releases/tag/v0.2.0? Otherwise we could have its own tag, releases/jsonschema/vXYZ, but I can't think of a case where we'll have the json schema version deviate.

znewman01:
> Is it sufficient to use an existing tag like https://github.com/sigstore/protobuf-specs/releases/tag/v0.2.0? Otherwise we could have its own tag, releases/jsonschema/vXYZ, but I can't think of a case where we'll have the json schema version deviate.

In the past, we liked versioning each set of bindings independently to handle changes to the *generation* process. E.g. if there was a bug in the protoc-jsonschema generator. I think we should default to that if possible.

haydentherapper:
SG, so we can just add a step to the release instructions to add a tag for `release/json/vX.Y.Z`, and we shouldn't need another release workflow.

woodruffw:
SGTM! I'll add a note to the release instructions.

haydentherapper (approved):

PR #114, opened 2023-07-28
common: message_digest is not required

woodruffw:
In most verification flows, this field will never be used.

This also tweaks the docs to clarify the field's role: it exists solely to **potentially** identify the artifact being verified.


protos/sigstore_common.proto
```diff
@@ -71,6 +71,8 @@ message HashOutput {
 // MessageSignature stores the computed signature over a message.
 message MessageSignature {
         // Message digest can be used to identify the artifact.
+        // Clients MUST NOT attempt to use this digest to verify the associated
+        // signature; it is intended solely for identification.
```
haydentherapper, line 75: Can we drop the required field behavior annotation?

woodruffw: My bad, I thought I did! Dropping.

haydentherapper (approved):

PR #121, opened 2023-08-03
jsonschema: container fix, updated compilation options

### LINKED ISSUE #120: jsonschema: Docker container dependency conflict
As of 1 hour ago on https://github.com/sigstore/protobuf-specs/actions/runs/5743015810/job/15566361392,

```
Step 1/4 : FROM alpine@sha256:82d1e9d7ed48a7523bdebc18cf6290bdb97b82302a8a9c27d4fe885949ea94d1
 ---> c1aabb73d233
Step 2/4 : RUN apk add --update protoc=3.21.12-r2 protobuf-dev=3.21.12-r2 go=1.20.5-r0 git=2.40.1-r0
 ---> Running in 7afa87ea9cd6
fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/community/x86_64/APKINDEX.tar.gz
ERROR: unable to select packages:
  go-1.20.7-r0:
    breaks: world[go=1.20.5-r0]
The command '/bin/sh -c apk add --update protoc=3.21.12-r2 protobuf-dev=3.21.12-r2 go=1.20.5-r0 git=2.40.1-r0' returned a non-zero code: 1
make: *** [Makefile:91: docker-image-jsonschema] Error 1
```

(CC @woodruffw , @kommendorkapten)

jleightcap:
#### Summary

Closes #120, towards #118


Dockerfile.jsonschema
```diff
@@ -1,6 +1,6 @@
 # 3.18.2
 FROM alpine@sha256:82d1e9d7ed48a7523bdebc18cf6290bdb97b82302a8a9c27d4fe885949ea94d1
-RUN apk add --update protoc=3.21.12-r2 protobuf-dev=3.21.12-r2 go=1.20.5-r0 git=2.40.1-r0
+RUN apk add --update protoc=3.21.12-r2 protobuf-dev=3.21.12-r2 go=1.20.7-r0 git=2.40.1-r0
```
woodruffw, line 3: CC @znewman01 for consensus, but I'd personally be okay with removing these pins now that we understand how they interact with the surrounding alpine image: the image itself has a "world" pin that effectively pins each of these dependencies.

(In other words, the pins here are extraneous and only cause trouble when the alpine image is updated automatically and they aren't.)

woodruffw (dismissed):
LGTM; one comment that I want additional feedback from the other maintainers on 🙂

Dockerfile.jsonschema
znewman01: Ahhh okay. That makes me a little nervous about auto-updates of the image itself. But I guess any big changes due to updates would get caught by the generation tests.

So I say go ahead and remove

woodruffw, line 3:
```suggestion
RUN apk add --update protoc protobuf-dev go git
```

PR #145, opened 2023-09-26
Add definition for Sigstore's DSSE signature extension

adityasaky:
#### Summary

Addresses https://github.com/sigstore/sig-clients/issues/9

This PR adds a definition for Sigstore's DSSE signature extension. The definition is a subset of the current VerificationMaterials message, omitting the option for a key ID in place of the cert chain. This is because the sigstore signature extension in a DSSE signature must only be used to communicate the intermediate certs and timestamp / tlog info.

https://github.com/secure-systems-lab/dsse/pull/61 introduces extensions to the DSSE specification.

#### Release Note

Define DSSE signature extension for Sigstore signatures.

#### Documentation

None.

### DIFF
--- a/protos/sigstore_bundle.proto
+++ b/protos/sigstore_bundle.proto
@@ -48,7 +48,14 @@ message TimestampVerificationData {
 }
 
 // VerificationMaterial captures details on the materials used to verify
-// signatures.
+// signatures. This message may be embedded in a DSSE envelope as a signature
+// extension. Specifically, the `ext` field of the extension will expect this
+// message when the signature extension is for Sigstore. This is identified by
+// the `kind` field in the extension, which must be set to
+// application/vnd.dev.sigstore.verificationmaterial;version=0.1 for Sigstore.
+// When used as a DSSE extension, if the `public_key` field is used to indicate
+// the key identifier, it MUST match the `keyid` field of the signature the
+// extension is attached to.
 message VerificationMaterial {
         oneof content {
                 dev.sigstore.common.v1.PublicKeyIdentifier public_key = 1 [(google.api.field_behavior) = REQUIRED];
--- a/gen/jsonschema/schemas/Bundle.schema.json
+++ b/gen/jsonschema/schemas/Bundle.schema.json
@@ -94,7 +94,7 @@
                 }
             ],
             "title": "Verification Material",
-            "description": "VerificationMaterial captures details on the materials used to verify signatures."
+            "description": "VerificationMaterial captures details on the materials used to verify signatures. This message may be embedded in a DSSE envelope as a signature extension. Specifically, the `ext` field of the extension will expect this message when the signature extension is for Sigstore. This is identified by the `kind` field in the extension, which must be set to application/vnd.dev.sigstore.verificationmaterial;version=0.1 for Sigstore. When used as a DSSE extension, if the `public_key` field is used to indicate the key identifier, it MUST match the `keyid` field of the signature the extension is attached to."
         },
         "dev.sigstore.common.v1.HashOutput": {
             "properties": {
--- a/gen/jsonschema/schemas/Input.schema.json
+++ b/gen/jsonschema/schemas/Input.schema.json
@@ -127,7 +127,7 @@
                 }
             ],
             "title": "Verification Material",
-            "description": "VerificationMaterial captures details on the materials used to verify signatures."
+            "description": "VerificationMaterial captures details on the materials used to verify signatures. This message may be embedded in a DSSE envelope as a signature extension. Specifically, the `ext` field of the extension will expect this message when the signature extension is for Sigstore. This is identified by the `kind` field in the extension, which must be set to application/vnd.dev.sigstore.verificationmaterial;version=0.1 for Sigstore. When used as a DSSE extension, if the `public_key` field is used to indicate the key identifier, it MUST match the `keyid` field of the signature the extension is attached to."
         },
         "dev.sigstore.common.v1.DistinguishedName": {
             "properties": {
--- a/gen/jsonschema/schemas/VerificationMaterial.schema.json
+++ b/gen/jsonschema/schemas/VerificationMaterial.schema.json
@@ -41,7 +41,7 @@
                 }
             ],
             "title": "Verification Material",
-            "description": "VerificationMaterial captures details on the materials used to verify signatures."
+            "description": "VerificationMaterial captures details on the materials used to verify signatures. This message may be embedded in a DSSE envelope as a signature extension. Specifically, the `ext` field of the extension will expect this message when the signature extension is for Sigstore. This is identified by the `kind` field in the extension, which must be set to application/vnd.dev.sigstore.verificationmaterial;version=0.1 for Sigstore. When used as a DSSE extension, if the `public_key` field is used to indicate the key identifier, it MUST match the `keyid` field of the signature the extension is attached to."
         },
         "dev.sigstore.bundle.v1.TimestampVerificationData": {
             "properties": {
--- a/gen/pb-go/bundle/v1/sigstore_bundle.pb.go
+++ b/gen/pb-go/bundle/v1/sigstore_bundle.pb.go
@@ -94,7 +94,14 @@ func (x *TimestampVerificationData) GetRfc3161Timestamps() []*v1.RFC3161SignedTi
 }
 
 // VerificationMaterial captures details on the materials used to verify
-// signatures.
+// signatures. This message may be embedded in a DSSE envelope as a signature
+// extension. Specifically, the `ext` field of the extension will expect this
+// message when the signature extension is for Sigstore. This is identified by
+// the `kind` field in the extension, which must be set to
+// application/vnd.dev.sigstore.verificationmaterial;version=0.1 for Sigstore.
+// When used as a DSSE extension, if the `public_key` field is used to indicate
+// the key identifier, it MUST match the `keyid` field of the signature the
+// extension is attached to.
 type VerificationMaterial struct {
 	state         protoimpl.MessageState
 	sizeCache     protoimpl.SizeCache
--- a/gen/pb-python/sigstore_protobuf_specs/dev/sigstore/bundle/v1/__init__.py
+++ b/gen/pb-python/sigstore_protobuf_specs/dev/sigstore/bundle/v1/__init__.py
@@ -19,9 +19,9 @@ class TimestampVerificationData(betterproto.Message):
     in the future.
     """
 
-    rfc3161_timestamps: List[
-        "__common_v1__.Rfc3161SignedTimestamp"
-    ] = betterproto.message_field(1)
+    rfc3161_timestamps: List["__common_v1__.Rfc3161SignedTimestamp"] = (
+        betterproto.message_field(1)
+    )
     """
     A list of RFC3161 signed timestamps provided by the user. This can be used
     when the entry has not been stored on a transparency log, or in conjunction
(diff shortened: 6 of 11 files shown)


haydentherapper:
cc @jku @loosebazooka @tomhennen too

protos/sigstore_bundle.proto
```diff
@@ -67,6 +67,21 @@ message VerificationMaterial {
         TimestampVerificationData timestamp_verification_data = 4;
 }
 
+// DSSESignatureExtension captures the verification materials to be embedded
+// with a Sigstore signature in a DSSE envelope. Specifically, this message is
+// meant to be used for the `ext` field in a signature extension. The `kind` for
+// the Sigstore DSSE extension MUST be
+// application/vnd.dev.sigstore.dsse+json;version=0.1. The certificate chain
+// included MUST NOT include the Sigstore root certificate.
```
loosebazooka, line 75: I think we say intermediates in other places? But I could be mistaken
```suggestion
// included MUST NOT include the Sigstore root certificate or trusted intermediates
```

protos/sigstore_bundle.proto
```diff
@@ -67,6 +67,22 @@ message VerificationMaterial {
         TimestampVerificationData timestamp_verification_data = 4;
 }
 
+// DSSESignatureExtension captures the verification materials to be embedded
+// with a Sigstore signature in a DSSE envelope. Specifically, this message is
+// meant to be used for the `ext` field in a signature extension. The `kind` for
+// the Sigstore DSSE extension MUST be
+// application/vnd.dev.sigstore.dsse+json;version=0.1. The certificate chain
```
adityasaky, line 74: Actually, I'm not sure we need `+json` here as this is never serialized independently.

protos/sigstore_bundle.proto
```diff
@@ -67,6 +67,22 @@ message VerificationMaterial {
...
+// included MUST NOT include the Sigstore root or trusted intermediate
+// certificates.
+message DSSESignatureExtension {
+        dev.sigstore.common.v1.X509CertificateChain x509_certificate_chain = 1 [(google.api.field_behavior) = REQUIRED];
```
haydentherapper, line 78: Should this be required? Should we support the case where only a public key is used, rather than a certificate? (in which case the envelope's keyID will suffice)

adityasaky: I'm trying to understand this case fully. Is the idea that the signature isn't issued via Fulcio but it's still necessary to verify both the signature and inclusion in rekor?

adityasaky: cc @mnm678

adityasaky: Right, understood. I'm reasoning about what the impact is on not having any required fields here. i.e., what it means for a verifier if an extension has only intermediate certificates? Should it reject because in the sigstore case we _need_ some timestamp data as well? What if there's only timestamp data? And so on. It's not directly linked to this specific thread but it got me thinking generally.

adityasaky: Is there a use case where we don't want one of tlog_entries and timestamp_verification_data? If neither is set, that indicates the extension may be used only for communicating arbitrary intermediate certs, which isn't the purpose. But as I understand it, a strict `oneof` is too strict as we can have both set?

haydentherapper: For the happy path for verification, we would expect:
- oneof(certificate chain, key ID as hint), where the chain is composed of an optional number of intermediates and a required leaf
- optionally `tlog_entries` - meaning that the client is in a private environment that doesn't need transparency
- optionally a timestamp 

The tricky case is if neither tlog_entires nor a timestamp is provided, the certificate can't be verified. So you also need at least one timestamp or `tlog_entries.inclusion_promise.signed_entry_timestamp` (which can't be encoded in protobuf logic)

kommendorkapten: The required pieces are described in the [sigstore client spec](https://docs.google.com/document/d/1kbhK2qyPPk8SLavHzYSDM8-Ueul9_oxIMVFuWMWKz0E/edit#heading=h.g11ovq2s1jxh), which defines the steps and requirements for a vanilla sigstore client, and those requirements are:

1. Signed with a X.509 certificate from Fulcio
2. Verify the timestamp of the signature creation with either (or both, but at least one!)
    1. A transparency log entry from Rekor
    2. A RFC3161 counter signature

For certain private deployments, the bundle can be used in more flexible ways, and I think we should be prescriptive in our language on how to use the bundle and include documentation that there SHOULD always be at least one transparency log entry _or_ one RFC3161 signed timestamp. Spending too much effort into enforcing certain behaviour into protobuf is not enough IMHO. The documentation should be clear on the expectations, as the canonical distribution format is JSON, a client may equally use a "regular" JSON parser, and so any constraints in the protobuf file is moot.

For private/public key pairs, there is already some redundancy in the current bundle definition. As an example, npm registry produces attestations in a DSSE envelope, and writes the key hint in both the bundle and the DSSE envelope. I don't think that's too bad, it's a small price to duplicate a key-hint to get consistency. 
Also remember that the hint is unauthenticated, verifiers should already have a set of trusted keys prepared prior to verification, so I would not put too much trust anyway in the hint.

mnm678: I agree with @kommendorkapten that this can be specified in the documentation rather than (or in addition to) the protobuf.

haydentherapper: To circle back to this, @kommendorkapten, are you saying you'd prefer the duplication of specifying a key hint in multiple places rather than maintaining this separate message?

adityasaky: > For the happy path for verification, we would expect:

>    oneof(certificate chain, key ID as hint), where the chain is composed of an optional number of intermediates and a required leaf
    optionally tlog_entries - meaning that the client is in a private environment that doesn't need transparency
    optionally a timestamp

> The tricky case is if neither tlog_entires nor a timestamp is provided, the certificate can't be verified. So you also need at least one timestamp or tlog_entries.inclusion_promise.signed_entry_timestamp (which can't be encoded in protobuf logic)

Okay, I think I understand this. In the case of DSSE where we have the key ID from the signature field already, this is simplified to:
* optionally the cert chain -> for signatures issued by sigstore
* timestamp information either through the timestamp field or from the tlog entries -> having just this is for signatures that may be signed by some arbitrary key pairs but the intent is to still verify it with rekor.

Is that a fair summary? I do agree having parsed @kommendorkapten's message that we have to specify in the docs rather than try to encode all of this logic in the proto definition.

> To circle back to this, @kommendorkapten, are you saying you'd prefer the duplication of specifying a key hint in multiple places rather than maintaining this separate message?

I'll leave this to the sigstore maintainers, if the only difference between the message proposed here is the key ID field. I'm not a fan of the repetition because as far as I can tell, it's also not really multiple distinct pointers to the key to use. That said, I also understand if the preference is to not define a new message which only drops the key ID field.

If the preference is to reuse verification materials, I can update this PR to merely add an explicit media type for it so the DSSE spec can refer to it.

kommendorkapten: > To circle back to this, @kommendorkapten, are you saying you'd prefer the duplication of specifying a key hint in multiple places rather than maintaining this separate message?

Not in particular, but with the current bundle layout, yes.

Today, the bundle contains (not all fields are listed):
* verification material
* content (`message_signature` or `dsse_enevelope`)

The verification material may contain a public key hint, and so does a DSSE envelope, but not a message signature. To avoid having complicated description that describes if the key hint in the verification material should have precedence over the DSSE envelope, or vice versa I would favour simplicity and so state that they should be equal, yet redundant. This would simplify any client code working with the bundle.

This also makes the DSSE envelope standalone, i.e removing the DSSE envelope from the bundle, we don't loose any information, and providing it to a "raw DSSE verifier" should still work. The counter-example would be that we document that the bundle should contain the key hint but not the DSSE envelope, that feels very wrong. 

Unrelated to this PR, I propose that we clarify this in the bundle, see this PR: https://github.com/sigstore/protobuf-specs/pull/148

Not that I don't think the current state is ideal, but we are layering a few formats on-top of each other, and some of them (like DSSE) are more prominent than others.

Also the canonical use-case for Sigstore is to _not_ rely on public key hints in the DSSE envelope -- it's to use X.509 certificates from Fulcio, so this issue is not expected to happen too often (but for npm's publish attestation it occurs).

adityasaky: Okay, I think I understand this from the context of handling the bundle, thanks!

For the DSSE extension, do you prefer using the new message here that drops the hint entirely or the existing `VerificationMaterials` message with documentation to indicate how the key ID must be checked?

adityasaky: I've updated this PR to reuse verification material after #148. Now, this PR adds a media type for verification material so that it can be used as `kind` in the signature extension.

kommendorkapten: > For the DSSE extension, do you prefer using the new message here that drops the hint entirely or the existing VerificationMaterials message with documentation to indicate how the key ID must be checked?

Yeah, I think so. So if the bundle is extracted from the DSSE envelope, it would be complete (except for the `content` as the composition is now the other way around).

kommendorkapten: Maybe clarify that the `content` is not available as that would be a recursive inclusion of the DSSE envelope. 

As content (`message_signature` or `dsee_envelope`) is marked as `REQUIRED` it could be clarified that a verifyer should re-assemble the bundle before using it?

adityasaky: @kommendorkapten as DSSE's signature extension embeds `VerificationMaterial`, `content` from `Bundle` is never encountered (i.e., there's no option to include it). Could you clarify?

>As content (message_signature or dsee_envelope) is marked as REQUIRED it could be clarified that a verifyer should re-assemble the bundle before using it?

I think this may be a slightly separate point. Do you mean the documentation should note that a DSSE envelope with the verification material sig ext could be recomposed to form a sigstore bundle?

kommendorkapten: Sorry @adityasaky I got it wrong. When I read through your updated PR I for some reason read it that you included the entire bundle in as an extension, please disregard my message above.

adityasaky: Understood, thanks for you review, I'm going to resolve this thread.

adityasaky:
I've re-emphasized that the public_key field must match the keyid in the dsse signature per #148. I've also cleaned up my commits etc.

trishankatdatadog (approved):

haydentherapper (approved):
I'm good with this! Thanks for all the discussion (and patience) on this. Tagging a few more maintainers: @woodruffw @kommendorkapten @mnm678 @bdehamer

haydentherapper:
@adityasaky Not sure why the tests didn't catch this, but can you run `make all` to update the comments in the generated files?

adityasaky:
Done!

haydentherapper (approved):

protos/sigstore_bundle.proto
```diff
@@ -48,7 +48,14 @@ message TimestampVerificationData {
...
 // VerificationMaterial captures details on the materials used to verify
-// signatures.
+// signatures. This message may be embedded in a DSSE envelope as a signature
+// extension. Specifically, the `ext` field of the extension will expect this
+// message when the signature extension is for Sigstore. This is identified by
+// the `kind` field in the extension, which must be set to
+// application/vnd.dev.sigstore.verificationmaterial;version=0.1 for Sigstore.
+// When used as a DSSE extension, if the `public_key` field is used to indicate
+// the key identifier, it MUST match the `keyid` field of the signature the
+// extension is attached to.
```
woodruffw, line 58: Just to make sure I'm not missing anything: this is the PR's only change, right? I don't see any protobuf definition changes 🙂

woodruffw (approved):
LGTM -- one clarifying comment!

(Given that this changes only the docs, I'm a little bit hazy on what this does, since the doc change doesn't seem actionable to a Sigstore client implementer. But that is possibly just an audience thing.)

bdehamer (approved):

protos/sigstore_bundle.proto
adityasaky: Yes, the doc is the only change. This PR originally proposed a new message that was a subset of verification materials, but the feedback was to just use verification materials directly. The doc change unblocks https://github.com/secure-systems-lab/dsse/pull/61 which introduces the sigstore DSSE extension.

> the doc change doesn't seem actionable to a Sigstore client implementer

WRT a sigstore client, it's a question of whether it will support a DSSE envelope that uses the signature extension as an input to signature verification. See: https://github.com/sigstore/sig-clients/issues/9 and the follow up https://github.com/sigstore/sig-clients/issues/9#issuecomment-1665979019.

kommendorkapten (approved):

PR #148, opened 2023-10-04
Clarify key hint

kommendorkapten:
#### Summary
Clarified how the use of (reduntant) key hint.
Key hint is both specified in the bundle's `verification material` but may also be specified inside the `content` if it'a DSSE envelope. This discussion started in this PR https://github.com/sigstore/protobuf-specs/pull/145

This PR favours simplicity over efficiency; that is; the key hint should be equal and duplicated to simplify any consumption of the bundle, as the value may appear in multiple places. The key hint (if used) is usually a short string so I argue it's not too bad.

#### Release Note
Clarified how to populate the key hint in the bundle, when the content is a DSSE envelope with a key hint too.

#### Documentation
This PR is only documentation.

### DIFF
--- a/gen/jsonschema/schemas/Bundle.schema.json
+++ b/gen/jsonschema/schemas/Bundle.schema.json
@@ -11,7 +11,7 @@
                 "verificationMaterial": {
                     "$ref": "#/definitions/dev.sigstore.bundle.v1.VerificationMaterial",
                     "additionalProperties": false,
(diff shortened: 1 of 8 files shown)


woodruffw (approved):
No objection to the language, although I'll repeat a previous general concern about identifier/value duplication: every time we duplicate a value (and expect a conforming client to check it for consistency), we introduce a possible error state (or arguably worse, an ambiguous state where a validation path exists but some clients may be looser or stricter than others). Maybe that's unavoidable in this case, but I think as a general design principle we should continue to deduplicate state wherever possible.

woodruffw:
IMO the current language is fine -- requiring that the two be exactly equal makes the most sense to me, I'm just griping about having to think about an additional error state :-)

PR #179, opened 2023-12-15
Added an option for a more generic observer time, and aligned with v0.2

kommendorkapten:
#### Summary
Based from the discussion https://github.com/sigstore/sigstore-go/issues/44

Currently there is not possible to define a policy that either a SET _or_ a signed timestamp should be used. This PR addresses that via the `ObserverTimestampOptions`. Also as v0.2 does prefer inclusion proofs over SET, this PR allows for option on using the SET or not, and when using the SET, the timestamp can be used during X.509 certificate verification.

#### Release Note
TBD

#### Documentation
N/A


protos/sigstore_verification.proto
```diff
@@ -72,6 +76,14 @@ message ArtifactVerificationOptions {
                 // Disable signed timestamp verification.
                 bool disable = 2;
         }
+        message ObserverTimestampOptions {
```
haydentherapper, line 79: I can ask this on sigstore-go too, but why was "Observer" chosen, over "Trusted" or "Verified" or just "TimestampOptions"?

protos/sigstore_verification.proto
```diff
@@ -54,6 +54,10 @@ message ArtifactVerificationOptions {
                 bool perform_online_verification = 2;
                 // Disable verification for transparency logs.
                 bool disable = 3;
+                // Verify SET timestamps indicates that the timestamp from
+                // the SET should be used when verifying the X.509
+                // certifiacte chain
+                bool verify_set_timestamp = 4;
```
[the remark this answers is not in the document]
haydentherapper: My last bullet point is wrong. I think `verify_set` must be true in order to count towards the ObserverTimestampOptions. If `verify_set=true` is set and only the TSA threshold is set up, then no threshold would be met (working as intended).

(tldr below is just me trying this out in a client)

As I've been prototyping this in sigstore-go, the only question I have is if it's sufficient to only set `WithTransparencyLog`. This means that the threshold for number of logs and timestamps must match, which I would prefer each should be explicitly configured.

I think the correct client approach would be to require when is used `WithTransparencyLog(threshold int, verifySET bool)`, also one of the following is required: `WithTSA(threshold)` or `WithObserverTimestamp(threshold)` or `WithInsecureNoTimestamp`. If `verifySET==true`, then the timestamps returned from verifying the log can count towards the `WithObserverTimestamp` threshold. If `WithTransparencyLog(1, true)` and `WithTSA(1)`, then SETs could be used to satisfy the log threshold, but not the timestamp threshold.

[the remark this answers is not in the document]
haydentherapper: OK, sorry I keep going back and forth, but I thought of one case this doesn't handle. How would you specify requiring X log proofs, 0 promises, and Y timestamps that are either SETs or TSA timestamps? This was the point of my last bullet point, some way to specify SETs shouldn't count towards.

So either:
* we need SETs to be their own struct to support this, or:
* a comment that `verify_set` only dictates whether or not SETs count towards the log threshold. We will always still try to verify SETs and extract timestamps, whose policy is dictated by ObserverTimestampOptions.

cc @steiza @phillmv 

Sorry for rambling on about this, hopefully this makes sense! Late Friday isn't the time to figure this out 😄

protos/sigstore_verification.proto
kommendorkapten: I just went with observer as that's what's used in https://github.com/sigstore/sigstore-go/issues/44

I have no real preference, or maybe I do prefer `TimestampOptions` as long as it's clear compared to `TimestampAuthorityOptions` and that makes me thing that `VerifiedTimestampOptions` might be the best.

I kept the comment a bit longer on purpose to include my reasoning, hope it makes sense.

protos/sigstore_verification.proto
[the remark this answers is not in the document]
kommendorkapten: As v0.2 has the SET as optional, and the proof as required. My thinking in this discussion was only around how the SET counts toward the timestamp threshold, not against the tlog threshold, and so:

> a comment that verify_set only dictates whether or not SETs count towards the log threshold. We will always still try to verify SETs and extract timestamps, whose policy is dictated by ObserverTimestampOptions.

would be sort of the opposite: _a comment that verify_set only dictates that it count towards the timestamp_ (if present)

This would mean that if the tlog does only contain a proof, this entry does not count up towards the timestamp threshold. I.e a tlog entry without SET but verified with `verify_set=true` would not fail, just not increase the counter towards timestamp threshold.

> How would you specify requiring X log proofs, 0 promises, and Y timestamps that are either SETs or TSA timestamps

Like this: 
WithTLog(x, true) + WithObserverTimestamp(y)

My reading:
WithTLog(x, true): We require x inclsuion proofs (required by v0.2 bundle), `verifySET=true` we trust the timestamp from the SET if included.
WithObserverTimestamp(y): we require at least y timestamps (as `verifySET=true`, the SET counts if present, if not then y TSA timestamps is required).

protos/sigstore_verification.proto
```diff
@@ -54,6 +54,12 @@ message ArtifactVerificationOptions {
                 bool perform_online_verification = 2;
                 // Disable verification for transparency logs.
                 bool disable = 3;
+                // Verify SET indicates that the timestamp from
+                // the SET should be used when verifying the X.509
+                // certifiacte chain. Note that inclusion promises (SETs)
+                // are optional in a bundle, and so requiring SET may render
+                // bundles invalid during verification if SET is not present.
+                bool verify_set = 4;
```
woodruffw, line 62: This is an extreme nitpick, but IMO the comment and the field name are slightly out of sync here: the comment says that the SET's embedded timestamp is *used to* fill in the X.509 chain's validation time, whereas the field name implies that the option enables verifying the SET itself.

Maybe `verify_time_from_set` or similar?

protos/sigstore_verification.proto
```diff
@@ -54,6 +54,12 @@ message ArtifactVerificationOptions {
                 bool perform_online_verification = 2;
                 // Disable verification for transparency logs.
                 bool disable = 3;
+                // Verify SET indicates that the timestamp from
+                // the SET should be used when verifying the X.509
+                // certifiacte chain. Note that inclusion promises (SETs)
```
woodruffw, line 59:
```suggestion
                // certificate chain. Note that inclusion promises (SETs)
```

protos/sigstore_verification.proto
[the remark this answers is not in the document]
haydentherapper: Good point about v0.2 requiring inclusion proof. Continuing discussion on sigstore-go.

protos/sigstore_verification.proto
kommendorkapten: Yeah, it was first called `verify_set_timestamp = 4;` but changed based on review. Let's keep this open until the PR in `sigstore-go` https://github.com/sigstore/sigstore-go/issues/44 is resolved, where this exact semantics is being discussed. I think we should align the naming between them.

kommendorkapten:
Updated to match this PR: https://github.com/sigstore/sigstore-go/pull/51

protos/sigstore_verification.proto
```diff
@@ -72,6 +72,20 @@ message ArtifactVerificationOptions {
...
                 bool disable = 2;
         }
+        message TlogIntegratedTimestampOptions{
+                // The number of integrated timestamps that are expected.
+                int32 threshold = 1;
+                // Disable integrated timestamp verification.
+                bool disable = 2;
+        }
+        message ObserverTimestampOptions {
+                // The number of external ovservers of the timestamp,
```
haydentherapper, line 82:
```suggestion
                // The number of external observers of the timestamp.
```

protos/sigstore_verification.proto
```diff
@@ -72,6 +72,20 @@ message ArtifactVerificationOptions {
...
+                // this is a union of RFC3161 signed timestamps, and
```
haydentherapper, line 83:
```suggestion
                // This is a union of RFC3161 signed timestamps, and
```

protos/sigstore_verification.proto
```diff
@@ -72,6 +72,20 @@ message ArtifactVerificationOptions {
...
+                // integrated timestamps from a transparency log
```
haydentherapper, line 84:
```suggestion
                // integrated timestamps from a transparency log,
                // though could include additional timestamp sources
                // in the future.
```

protos/sigstore_verification.proto
```diff
@@ -72,6 +72,20 @@ message ArtifactVerificationOptions {
...
+                int32 threshold = 1;
+                // Disable signed timestamp verification.
```
haydentherapper, line 86:
```suggestion
                // Disable observer timestamp verification.
```

protos/sigstore_verification.proto
```diff
@@ -94,6 +108,7 @@ message ArtifactVerificationOptions {
         // Threshold: 1
         // Online verification: false
         // Disable: false
+        // Verify SET Timestamps: true
```
haydentherapper, line 111:
```suggestion
```

haydentherapper: Since this option no longer controls SET timestamp verification

protos/sigstore_verification.proto
```diff
@@ -106,6 +121,11 @@ message ArtifactVerificationOptions {
         // Threshold: 1
         // Disable: false
         optional TimestampAuthorityOptions tsa_options = 5;
```
haydentherapper, line 123: Also need `optional TlogIntegratedTimestampOptions integrated_ts_options`

haydentherapper (commented):
LGTM, just a few comment updates

protos/sigstore_verification.proto
kommendorkapten: Good catch!

kommendorkapten:
As part of adding the `ObserverTimestampOptions`, I modified the default verification parameters to be more user-friendly. The default option is now to require one `ObserverTimestamp` (that is, one from a tlog or from a RFC3161), if a user would like to control this better they can override this. I think this makes the default options easier to reason about, and aligns with the example `sigstore-go` binary: https://github.com/sigstore/sigstore-go/blob/1195bd5938daaf53eb2c011608235880e8a8524c/cmd/sigstore-go/main.go#L59

haydentherapper (approved):
Looks great! Thank you again for the great discussion on this.

Issue #189, opened 2024-01-09
Add a registry of supported signature suites/hash algorithms


woodruffw:
Opening as a tracking issue; CCing @tetsuo-cpp and @ret2libc.

From initial discussion, `protobuf-specs` is probably the right place for this registry to live. It should probably be both a Markdown file and a `.proto` (containing enums/string constants for the cipher suite names, maybe?)

Blocked on completion of the relevant parts of the [Configurable Crypto Algorithms document](https://docs.google.com/document/d/18vTKFvTQdRt3OGz6Qd1xf04o-hugRYSup-1EAOWn7MQ/edit?usp=sharing).

woodruffw:
Some relevant discussion items: 

* How "detailed" should the protobuf version of the registry be? In other words: should it list things like expected purposes (signing only, verifying only, both), or just be a "bare" list of every algorithm and suite that a Sigstore client might reasonably need to support?

tetsuo-cpp:
This is addressed in #194, we can close this now.

Issue #183, opened 2023-12-20
trustroot: Should OAuth issuer(s) be included?


jku:
This is not a fully defined feature request but I wanted to write this down before holidays...

I was testing if a sigstore client (sigstore-python) _really_ can choose the "sigstore instance" purely based on trusted_root.json: verification seems to work pretty well without any hard coded urls etc, but signing still has a an issue or at least a usability problem: A signing client can use trusted_root.json to select the Fulcio instance it uses, but it also has to choose the OAuth issuer (dex) URL and for things to work that issuer needs to be accepted by the Fulcio instance.

Should the `sigstore_trustroot` `CertificateAuthority` message include a parameter that describes the accepted OAuth issuers that this Fulcio instance claims to accept? Something like `repeated string oath_issuers` or even just a `string oath_issuer` -- the idea isn't necessarily limit the fulcio instance to just these issuers but help clients choose an issuer that is likely to work with the fulcio instance.

kommendorkapten:
That is an interesting thought! So far I have mostly considered `trusted_root.json` from the point of a verifier. 

If I get your question correct, we would list the issuers next to each CA, as an example the Public Good Instance Fulcio would countain `[github.com, google.com, microsoft.com]`? If say google creates a private sigstore instance that only accepts `[google.com]`, would that complicate it for the client? I.e do we expect the list of issuers to _always_ be disjoint?

The second question, would this imply that we need to sign and publish a new `trusted_root.json` if a new issuer is added to a Fulcio instance?

jku:
> If I get your question correct, we would list the issuers next to each CA, as an example the Public Good Instance Fulcio would countain [github.com, google.com, microsoft.com]?

otherwise yes, except this is about the non-federated issuer (dex). Correct me if I'm wrong but I believe both production and staging fulcio only accept a single issuer for "actual users" (`https://oauth2.sigstore.dev/auth` and `https://oauth2.sigstage.dev/auth` respectively)

woodruffw:
Registering my support for this idea: I've also mostly been thinking about the trusted root from the verifier's perspective, but (IMO) it makes sense to allow signers to use it to convey their configuration/trusted state as well. 

[...]

Issue #66, opened 2023-02-01
Define Bundle Verification Result 


bdehamer:
We already have message types defined that describe the inputs to the verification process (`Bundle`, `TrustedRoot`, `ArtifactVerificationOptions`) so it seems reasonable to also define a standardized verification response. This will give clients a standard way to convey more specific verification errors (and make it easier to compare the verification results across the different client implementations).

I'm not married to this particular design, but offering it as a starting point for refinement:
```protobuf
enum VerificationStep {
        // Verify the signature w/ the leaf certificate in the chain
        SIGNATURE = 1;
        // Verify the certificate chain back to the root of the identity service
        CERTIFICATE_CHAIN = 2;
        // Verify the SET of the transparency log entry
        TRANSPARENCY_LOG = 3;
        // Verify the signed timestamp is valid and trusted
        TIMESTAMP_AUTHORITY = 4;
        // Verify the signature timestamp (either from tlog SET or timestamp authority) is within certificate's validity period
        SIGNATURE_TIMESTAMP = 5;
}

message VerificationResult {
        bool success = 1;
        optional string message = 2;
        optional VerificationStep failed_step = 3;
}
```

If we go with this idea of having some sort of enum to identity the specific verification step which failed we could go even more granular (e.g. `CERTIFICATE_CHAIN_SCT`, `TRANSPARENCY_LOG_INCLUSION_PROOF`, etc) . . . not clear to me what the most useful level of detail is gonna be.

Another thing which occurs to me is that it might be helpful for the `VerificationResult` to surface the Fulcio certificate extensions (in the case where the verification is successful). The extensions are probably the next things any client is gonna read after verifying the bundle and this could save them the trouble of fishing them out of the certificate.

CC @kommendorkapten @asraa

asraa:
> Another thing which occurs to me is that it might be helpful for the VerificationResult to surface the Fulcio certificate extensions (in the case where the verification is successful). 

+1 @vaikas, any other things useful or interesting?

My own thoughts about the error struct:
* I think there should be different classes of errors. I think you've described something like "verification errors", but I think there are other "types" or classes, like "validation errors" or "parsing errors" to represent malformed content (or do we assume that these are already checked for?), "internal errors" (like if you request an online proof and it times out), and maybe others
* I think "Signature timestamp" is actually a class of "Certificate errors" or certificate validity. In the broader context of all types of PKI, it may not be necessary that a signature timestamp must be checked, as if you use a key.

[...]

PR #191, opened 2024-01-10
gen, protos: 0.3, single cert

### LINKED ISSUE #190: Proposal: Specify only leaf cert in bundle
Background: The Sigstore bundle includes the verification material for a signature. For a signature generated with a key, it includes a public key fingerprint. When generated with a certificate, it currently can include either just the leaf certificate, or the leaf and its chain. The chain should only include intermediates, although there are bundles that were generated with roots too (resulting in [this comment](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L177). 

The [X509CertificateChain](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L172) message is currently used in two places: In the verification material for a bundle, or in the trust root. The comments on the message are mostly relevant to the former case.

As discussed in sig-clients, we would like to propose removing support for specifying a chain and require that the chain be distributed out of band or in the trust root. Including the chain in the bundle is useful for classical PKI cases where an intermediate is untrusted and only used for chain building. However the bundle wasn't designed with private deployments in mind. We do still want to support those cases, but want to be intentional in doing so.

One way we could this is to avoid breaking changes is to add an [additional field to the oneof in VerificationMaterial](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto#L60-L63), specifying a single certificate. Clients could then choose whether to support a BYO PKI case using `x509_certificate_chain`, or only support a single `certificate`. We could also do this in a breaking way and drop `x509_certificate_chain`.

Clients for thoughts: @loosebazooka @woodruffw @kommendorkapten @bdehamer @codysoyland

woodruffw:
This adds a new `oneof` variant to `VerificationMaterials.contents` that specifically encodes a single X.509 leaf certificate. This ratchets down (soft) invariants that were previously documented on the `x509_certificate_chain` member.

In the process, it also bumps the bundle version (`0.3`) and moves some documentation around.

Closes #190.


protos/sigstore_bundle.proto
```diff
@@ -57,9 +57,38 @@ message TimestampVerificationData {
...
+        // When used with the Public Good Instance (PGI) of Sigstore, form (1) MUST NOT be used,
+        // regardless of bundle version.
+        //
+        // When used in a `0.1` or `0.2` bundle with the PGI, form (2) MUST be used.
+        //
+        // When used in a `0.3` bundle with the PGI, form (3) MUST be used.
         oneof content {
                 dev.sigstore.common.v1.PublicKeyIdentifier public_key = 1 [(google.api.field_behavior) = REQUIRED];
                 dev.sigstore.common.v1.X509CertificateChain x509_certificate_chain = 2 [(google.api.field_behavior) = REQUIRED];
+                dev.sigstore.common.v1.X509Certificate certificate = 5 [(google.api.field_behavior) = REQUIRED];
```
woodruffw, line 91: Flagging: I numbered this as `5`, since we didn't reserve any additional `oneof` variants here. I don't understand protobuf well enough to be certain, but I *think* this preserves the wire format on older bundles 🙂

protos/sigstore_common.proto
```diff
@@ -168,25 +168,17 @@ message SubjectAlternativeName {
         }
 }
 
-// A chain of X.509 certificates.
+// A collection of X.509 certificates.
+//
+// NOTE: "Chain" is a misnomer in this context, since there is no one true certificate chain
+// in most PKI schemes. This message should be treated as a generic collection of certificates
+// for path construction.
```
woodruffw, line 175: Flagging: I added this `NOTE` based on the discussion we had in the clients call -- it might be overkill, though.

protos/sigstore_bundle.proto
haydentherapper: Yep, 5 is correct https://protobuf.dev/programming-guides/proto3/#oneof:~:text=Field%20numbers%20for%20oneof%20fields%20must%20be%20unique%20within%20the%20enclosing%20message.

protos/sigstore_bundle.proto
```diff
@@ -32,8 +32,8 @@ option ruby_package = "Sigstore::Bundle::V1";
 // The primary message ('Bundle') MUST be versioned, by populating the
 // 'media_type' field. Semver-ish (only major/minor versions) scheme MUST
 // be used. The current version as specified by this file is:
-// application/vnd.dev.sigstore.bundle+json;version=0.2
-// The semantic version is thus '0.2'.
+// application/vnd.dev.sigstore.bundle+json;version=0.3
```
haydentherapper, line 35: Just pressure testing - Do we want a new version at this point? I think yes. We could add the field without a new version to avoid this being a breaking change, but then the field would have to be optional.

protos/sigstore_bundle.proto
```diff
@@ -57,9 +57,38 @@ message TimestampVerificationData {
...
+        // 2. A sequence of one or more X.509 certificates, of which the first member
+        //    MUST be a leaf certificate conveying the signing key.
+        //
+        //    Signers MUST NOT include root CA certificates in bundles, and SHOULD NOT
+        //    include intermediate CA certificates that appear in an independent root of trust
+        //    (such as the Public Good Instance's trusted root).
+        //
+        //    Verifiers MUST validate the chain carefully to ensure that it chains up
+        //    to a CA certificate that they independently trust. Verifiers SHOULD
+        //    handle old or non-complying bundles that have additional intermediate and/or
```
haydentherapper, line 76: Should we define "handle" more precisely? Something about ignoring root CA certificates if present? As for intermediates, that's currently up to the client. Cosign is the outlier in that it'll read in any intermediates provided. Maybe we mention something about clients providing a separate path for PGI vs private instances?

protos/sigstore_bundle.proto
```diff
@@ -57,9 +57,38 @@ message TimestampVerificationData {
...
+        //    root CA certificates.
+        //
+        // 3. A single X.509 certificate, which MUST be a leaf certificate conveying
+        //    the signing key.
+        //
+        // When used with the Public Good Instance (PGI) of Sigstore, form (1) MUST NOT be used,
```
haydentherapper, line 82: Form 1 should be allowed for self-managed keys. You can use the public instance of Rekor without the public instance of Fulcio. I'm assuming this comment is about using Fulcio? Is there a way we can convey that if so?

protos/sigstore_common.proto
haydentherapper: It's not just used for path construction though, also for providing a root CA certificate for TUF use cases.

Maybe add proposed different use cases?

protos/sigstore_common.proto
```diff
@@ -168,25 +168,17 @@ message SubjectAlternativeName {
...
-        // Verifiers MAY enforce additional constraints, such as requiring that
-        // all intermediate CA certificates appear in an independent root of
-        // trust.
+        // One or more DER-encoded certificates.
         //
-        // Verifiers SHOULD handle old or non-complying bundles that have
-        // additional intermediate/root CA certificates.
+        // In some contexts (such as `VerificationMaterial.certificate`), this sequence
+        // has an imposed order. Unless explicitly specified, there is otherwise no
+        // guaranteed order.
```
haydentherapper, line 181: It's only a comment so we can't enforce easily, but should we be more prescriptive regarding order? Something like "index n+1 SHOULD be the issuing certificate for index n"?

protos/sigstore_bundle.proto
woodruffw: Yep, I think so as well -- IMO making the field non-optional would be confusing in context, since the other two `content` variants are marked as required.

(But maybe there are other good reasons to not bump the version here? Other client maintainers should probably chime in 🙂)

protos/sigstore_bundle.proto
woodruffw: Yeah, good catch -- I'll make it explicit that verifiers SHOULD handle these by ignoring any superfluous CA certificates (either by simply ignoring them entirely, or by ensuring that their chain building process treats them as untrusted for the purposes of verification). If I understand correctly, that'll be consistent with what `cosign` does 🙂

protos/sigstore_common.proto
woodruffw: We could do that, although there may be (I really hope nobody does this, but there's nothing stopping it) cross-issuing PKI topologies like this one:

```
RCA_A -> ICA_A'
RCA_B -> ICA_A''
ICA_A -> leaf
```

(i.e. `ICA_A` is a single logical CA with two separate CA certs issued by different root CAs, either of which can be a valid issuer for `leaf`).

In this topology, `N+1` could be either `ICA_A'` or `ICA_A''`, and both could be listed in the "chain" for completeness purposes. So it's not clear which to order first 🙂 

(In practice, maybe this doesn't matter since `SHOULD` isn't `MUST`?)

protos/sigstore_bundle.proto
woodruffw: Yeah, I'll clarify that this is about using Fulcio/"keyless" flows.

protos/sigstore_bundle.proto
kommendorkapten: I would think that a breaking change is desired _if_ we want clients to actually adopt the usage of the new `certificate`. If not, clients would need to write the certificate into both `x509_certificate_chain` and `certificate`, and I'm not sure that's desirable? 

I think we can roll this out quite easy by describing the change more of a clarification, and as part of that moving the certificate to a new field and making it explicit. This would of course still require clients to discriminate on the version, and this is less optimal for verifiers that would need to be compliant with multiple versions.

protos/sigstore_common.proto
kommendorkapten: I agree with @haydentherapper on being more prescriptive here. And as @woodruffw points out, if someone is doing a multi-rooted tree, they are are allowed as it's a `SHOULD`, and also, the verifier would probably be aware of this (i.e have both `RCA_A` and `RCA_B`) and should so be able to construct a chain regardless.

gen/jsonschema/schemas/Bundle.schema.json
```diff
@@ -65,6 +65,10 @@
                     "$ref": "#/definitions/dev.sigstore.common.v1.X509CertificateChain",
                     "additionalProperties": false
                 },
+                "certificate": {
```
loosebazooka, line 68: would x509certificate be more consistent or is it just kind of redundant?

I guess this naming is in line with X509CertificateChain.certificates?

protos/sigstore_common.proto
woodruffw: Sounds good, I'll make it more prescriptive!

gen/jsonschema/schemas/Bundle.schema.json
woodruffw: Yeah, I was thinking the same. `x509_certificate` would be consistent with the other variant, but it feels a bit redundant. I have no strong opinions here 🙂

protos/sigstore_common.proto
loosebazooka: Is there some situation in which we're expected to parse out this multirooted chain?

gen/jsonschema/schemas/Bundle.schema.json
loosebazooka: just leave as is unless @haydentherapper or @kommendorkapten chime in?

protos/sigstore_common.proto
woodruffw: 6d2a288 strengthens the prescription in the bundle case, PTAL 🙂

woodruffw: > Is there some situation in which we're expected to parse out this multirooted chain?

This could show up in a BYO PKI scenario where the root of trust contains multiple separat trust anchors, but I think it's pretty unlikely -- I mostly produced it to illustrate that the chain "order" is ill-defined outside of basic PKIs.

protos/sigstore_bundle.proto
woodruffw: > I would think that a breaking change is desired _if_ we want clients to actually adopt the usage of the new `certificate`. If not, clients would need to write the certificate into both `x509_certificate_chain` and `certificate`, and I'm not sure that's desirable?

My 0.02c would be for encouraging clients to adopt the `certificate` form -- the language in `VerificationMaterial` currently says that adopting the `0.3` bundle for keyless signing + PGI means that clients `MUST` use the `certificate` field instead of `x509_certificate_chain`. 

But yeah, this does make things a bit of a mess for verifying clients...

gen/jsonschema/schemas/Bundle.schema.json
haydentherapper: What if we add support for SSH certificates? 😄 

No but really, leaving off `x509` is fine with me. We probably should have named the other field `certificate_chain` as x509 is a bit too in the weeds.

protos/sigstore_bundle.proto
haydentherapper: For clients that wish to only support PGI, we could say just take the first cert of `cert-chain` or `certificate`.

kommendorkapten: I think the current wording is good 👍 . My comment was a general about clients that would need to deal with multiple versions, not how the fields are used to day. I.e that regardless what we do, clients have to pay the price :) This is of course not specific to this spec, it's a general statement of concurrently supporting multiple versions.

protos/sigstore_common.proto
```diff
@@ -168,25 +168,17 @@ message SubjectAlternativeName {
...
-        // up to a root CA certificate that they trust, regardless of whether
-        // the chain includes additional intermediate/root CA certificates.
-        // Verifiers MAY enforce additional constraints, such as requiring that
-        // all intermediate CA certificates appear in an independent root of
-        // trust.
+        // One or more DER-encoded certificates.
         //
-        // Verifiers SHOULD handle old or non-complying bundles that have
-        // additional intermediate/root CA certificates.
+        // In some contexts (such as `VerificationMaterial.certificate`), this sequence
```
kommendorkapten, line 179:
```suggestion
        // In some contexts (such as `VerificationMaterial.x509_certificate_chain`), this sequence
```

I believe this is what you meant? (Leaf is the first, then any intermediates.)

protos/sigstore_bundle.proto
woodruffw: Yeah, further version proliferation isn't ideal.

In the interest of keeping the number of version iterations to a minimum, maybe we should roll this up with a few other changes that we expect to occur in a new version? In particular, I'm thinking of #189, #183, #66, and maybe some others.

protos/sigstore_common.proto
woodruffw: Yes, thanks!

protos/sigstore_bundle.proto
kommendorkapten: Yes, maybe it's time we plan for a next release, i.e bulk more things together as you mentioned. Don't think there is any urgency around getting `v0.3` out?

kommendorkapten (dismissed):
Thanks for doing this 🚀

protos/sigstore_bundle.proto
woodruffw: No extreme urgency, although I would definitely like to get #189 merged sooner rather than later 🙂 -- there's a lot of work that isn't strictly blocked on it, but will probably become a client interop challenge unless we eventually enumerate the various sigs/hashes that can occur on the PGI.

haydentherapper (approved):
Nice work!

woodruffw:
In terms of procedure, should we merge here and let `0.3` live on `main`? I figure that's fine since it won't be tagged or released anywhere yet, but just confirming 🙂

kommendorkapten (approved):
Agree, merge this and let `0.3` live at `main` until we cut a release.

PR #199, opened 2024-01-19
Add RSA variants, experimental LMS and LM-OTS to algorithm registry

woodruffw:
Opening this up to get a discussion started! I suspect this will require a decent amount of discussion and review before it can be merged.

Some particular points worth highlighting:

* How should we communicate this? In particular, is it this repo/doc's responsibility to explain the complexity of hash-based schemes (especially non-OT ones)?
* Should we make parameter recommendations? Making recommendations for LM-OTS is straightforward (we probably want `LMOTS_SHA256_N32_W8` almost always), but LMS is harder to make general purpose recommendations for (since the different performance/size tradeoffs will vary for different signer setups).

CC @tetsuo-cpp @ret2libc for visibility


docs/algorithm-registry.md
```diff
@@ -15,16 +15,30 @@ specification for the design rationale for this registry.
...
+|           | rsa-sign-pkcs1-3072-sha256 | sign/verify |                                                                                  |
+|           | rsa-sign-pkcs1-4096-sha256 | sign/verify |                                                                                  |
+| ECDSA     | ecdsa-sha2-256-nistp256    | sign/verify |                                                                                  |
+|           | ecdsa-sha2-384-nistp384    | sign/verify |                                                                                  |
+|           | ecdsa-sha2-512-nistp521    | sign/verify |                                                                                  |
+| EdDSA     | ed25519                    | sign/verify |                                                                                  |
+|           | ed25519-ph                 | sign/verify | Recommended only for `hashedrekord`.                                             |
+| LMS       | lms-sha256                 | sign/verify | Stateful; signer selects the `H` parameter. Not recommended for keyless signing. |
```
woodruffw, line 28: Flagging for review: I added "not recommended for keyless signing" because the chief advantage of keyless signing is the ability to use arbitrary numbers of identity-bound keys, which is the ideal scenario for LM-OTS.

haydentherapper:
Will take a longer look next week!

A few quick thoughts:

- What is the intended use case of LMS? LMS-OTS is fairly clear, this is to be used with the default ephemeral signing code path (though maybe it'd be worth explicitly stating that this should be selected only for this case). When would we recommend someone use LMS? For self-managed keys where the signer keeps track of usage? 
- "is it this repo/doc's responsibility to explain the complexity of hash-based schemes (especially non-OT ones)" - This is a great question. I want to avoid as many footguns as possible. If LMS-OTS has a clear, safe use case, then we should constrain it to only that use case (in practice, this will just be in a comment, but maybe it can be enforced by disallowing the algorithm in services, or at least requiring some flag-based acknowledgement of the risks).
- For parameter selection, has anyone given suggestions so far? I can ask Google's ISE Crypto team. Has NIST? 

Another thought, should we put PQ algs into a separate experimental enum which has no guarantees with respect to breaking changes? That may also signal these are a work in progress.

woodruffw:
> * What is the intended use case of LMS? LMS-OTS is fairly clear, this is to be used with the default ephemeral signing code path (though maybe it'd be worth explicitly stating that this should be selected only for this case). When would we recommend someone use LMS? For self-managed keys where the signer keeps track of usage?

That's my understanding of the use case, but I'm just the lowly implementer 🙂. @tpletcher-hpe might be able to offer some additional context (here or on Slack; I can set that chat up).

>  If LMS-OTS has a clear, safe use case, then we should constrain it to only that use case (in practice, this will just be in a comment, but maybe it can be enforced by disallowing the algorithm in services, or at least requiring some flag-based acknowledgement of the risks).

This makes sense to me -- the clear, safe case for LM-OTS is keyless signing, which should be relatively misuse-resistant so long as the Sigstore client always generates a new keypair per signing operation.


> * For parameter selection, has anyone given suggestions so far? I can ask Google's ISE Crypto team. Has NIST?

RFC 8554 has some parameter recommendations, including "top-level" recommendations for HSS (constructed on LMS) based on different tree depths and signing volume/longevity assumptions. All of those recommendations have reasonable security margins, but unfortunately aren't easy to generalize (since some clients may only need a key-tree that's valid for a few minutes, while another might need one that's valid for decades).

> Another thought, should we put PQ algs into a separate experimental enum which has no guarantees with respect to breaking changes? That may also signal these are a work in progress.

That sounds reasonable to me. @tetsuo-cpp and @ret2libc: do you think having these in a separate enum would pose an integration challenge?

docs/algorithm-registry.md
kommendorkapten: I seemed to miss the original PR that added this file and `KnownSignatureAlgorithm`, but what is the difference from the [PublicKeyDetails](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L52)? 

It seems a bit redundant now? And they are not really in sync either.

We did start out to separate algorithm and encoding to different messages, but decided to merge them, to avoid situation where invalid combinations was used to make it easier for clients with less failure modes.

ret2libc:
> That sounds reasonable to me. @tetsuo-cpp and @ret2libc: do you think having these in a separate enum would pose an integration challenge?

I think it would be ideal to treat them as the same type, otherwise integration might become harder. If we can keep them separate while still treating them the same type on the Go-side (and other languages as well, actually), then I'm ok.

woodruffw:
> I think it would be ideal to treat them as the same type, otherwise integration might become harder. If we can keep them separate while still treating them the same type on the Go-side (and other languages as well, actually), then I'm ok.

Hmm, this might be hard without a layer of wrapper types in between. Given that enum variants are "cheap" and that nothing else uses `KnownSignatureAlgorithm` yet, maybe we could add these to the same enum but explicitly prefix them as `EXPERIMENTAL_...` to emphasize that they aren't considered fully recommended yet?

Thoughts @haydentherapper?

tetsuo-cpp:
> Given that enum variants are "cheap" and that nothing else uses KnownSignatureAlgorithm yet, maybe we could add these to the same enum but explicitly prefix them as EXPERIMENTAL_... to emphasize that they aren't considered fully recommended yet?

I'd prefer that idea also. If they're captured in separate enums, we'll have to wrap the enums behind an `interface` or something which will be messy in my opinion.

haydentherapper:
That sounds good with me.

kommendorkapten:
Yes, `EXPERIMENTAL` prefix is the better option. When the algorithms are stable, we can mark the experimental field with `[deprecated = true]` and create a new entry for the stable one. This gives us a neat transition period.

kommendorkapten:
But I still want to raise the question of how `KnownSignatureAlgorithm` should play with `PublicKeyDetails`.

The `PublicKeyDetails` today conflates two things, signature algorithm and encoding (on purpose to minimize footguns). 
The `PublicKeyDetails` are used for rekor and the ct log as an example. 

Wouldn't it make sense to only use that enum to capture the use-case here too (i.e add new experimental algorithms)? From a clients side perspective I'm kind of scratching my head a little, what's the intent of `KnownSignatureAlgorithms` when the algorithm can not be communicated to the client, and it can signal an algorithm we can't properly provide a public key for? (we can technically of course, but not as the protos are defined today).

protos/sigstore_common.proto
```diff
@@ -79,6 +79,13 @@ enum PublicKeyDetails {
         // Ed 25519
         PKIX_ED25519 = 7; // See RFC8032
         PKIX_ED25519_PH = 8;
+
+        // **EXPERIMENTAL**: LMS and LM-OTS variants.
+        EXPERIMENTAL_LMS_SHA256 = 14;
+        EXPERIMENTAL_LMOTS_SHA256 = 15;
+
+        // Reserved for future additions of public key/signature algorithm types.
+        reserved 16 to 50;
```
woodruffw, line 88: NB: I've added a reserved range here, similar to some of the other messages we have. I figure this will make things a little more flexible on the wire format side, although I'm not sure if it matters in the JSON serialization case.

woodruffw:
Rewritten to extend `PublicKeyDetails` instead, PTAL @haydentherapper @kommendorkapten!

protos/sigstore_common.proto
```diff
@@ -69,6 +69,9 @@ enum PublicKeyDetails {
         PKIX_RSA_PKCS1_2048_SHA256 = 9;
         PKIX_RSA_PKCS1_3072_SHA256 = 10;
         PKIX_RSA_PKCS1_4096_SHA256 = 11;
+        RSA_PSS_2048_SHA256 = 16;
+        RSA_PSS_3072_SHA256 = 17;
+        RSA_PSS_4096_SHA256 = 18;
```
woodruffw, line 74: From https://github.com/sigstore/protobuf-specs/pull/212#discussion_r1476477104: maybe we want to explicitly prefix these with `PKIX_`, emphasizing that they're encoding according to RFC 4055? Thoughts?

haydentherapper (approved):
Thanks for the updates! 

Can you update the PR title to note the RSA scheme additions?

protos/sigstore_common.proto
```diff
@@ -66,9 +66,22 @@ enum PublicKeyDetails {
...
+        PKIX_RSA_PSS_3072_SHA256 = 17;
+        PKIX_RSA_PSS_4096_SHA256 = 18;
+        // RSA public key in PKCS#1 format, PKCS#1v1.5 signature
+        PKCS1_RSA_PKCS1V15_2048_SHA256 = 19;
+        PKCS1_RSA_PKCS1V15_3072_SHA256 = 20;
+        PKCS1_RSA_PKCS1V15_4096_SHA256 = 21;
+        // RSA public key in PKCS#1 format, RSASSA-PSS signature
+        PKCS1_RSA_PSS_2048_SHA256 = 22; // See RFC4055
+        PKCS1_RSA_PSS_3072_SHA256 = 23;
+        PKCS1_RSA_PSS_4096_SHA256 = 24;
```
woodruffw, line 84: Here's the full linearization of supported RSA key format + signature format options.

As a random thought: given that the PKIX encoding is just an encapsulation of the PKCS1 encoding for RSA public keys, maybe it makes sense to encourage the ecosystem as a whole to uniformly use the PKIX encoding? In other words: rather than explicitly enumerating the `PKCS1_RSA_*` variants, maybe we could say that they're subsumed under the deprecated `PKCS1_RSA_PKCS1V5` variant and that implementations *should* transform their public keys into PKIX instead?

(I don't know if this is a good idea or not, it just came to mind as an option if we want to reduce the enum size here 🙂)

woodruffw:
> Can you update the PR title to note the RSA scheme additions?

Done!

protos/sigstore_common.proto
haydentherapper: For the public instance, we will only use PKIX. PKCS1 would cover private deployments. Do you know how much support there is for PKCS1 encodings in other languages? As in, if the only sigstore client that supports pkcs1 encoded rsa keys is cosign, maybe we drop these for now?

woodruffw: >  Do you know how much support there is for PKCS1 encodings in other languages? As in, if the only sigstore client that supports pkcs1 encoded rsa keys is cosign, maybe we drop these for now?

Not sure about JS or Java, but we don't support PKCS1-encoded keys in `sigstore-python`. CC @bdehamer @loosebazooka @kommendorkapten for additional datapoints here 🙂

kommendorkapten: I would to ignore PKCS1 encoded keys (that's why I didn't add them, only kept the old definitions as `deprecated`). I'll look into JS.

kommendorkapten: sigstore-js only supports PKIX keys now: https://github.com/sigstore/sigstore-js/blob/ed19022f2817c8444cc343984609102d0324fe4f/packages/core/src/crypto.ts#L25

Side note is that nodejs documentation also recommends using PKIX for persisting public keys for long-term storage.

That said it would be possible to enable usage of PKCS1 encoded keys in sigstore-js, but as mentioned earlier, I rather not.

kommendorkapten:
Thanks, good to see this. My preference is to remove the PCS1 encoding for RSA, then we are ready to merge.

protos/sigstore_common.proto
woodruffw: Makes sense! Given that, I think we can entirely drop the PKCS1 variants here (and re-add later, if needed).

woodruffw:
> Thanks, good to see this. My preference is to remove the PCS1 encoding for RSA, then we are ready to merge.

Done!

kommendorkapten (approved):
🚀

PR #207, opened 2024-01-31
pb-rust: codegen via `prost` + `prost-reflect`

tnytown:
#### Summary

Generate Protobuf structs via `prost`, handling the Protobuf JSON mapping with `prost-reflect`.

This approach has a few benefits:
- Handles `base64` encoding of the `bytes` type automatically.
- Converts `i64` correctly.
- Better struct naming and module structure.

Unfortunately, `prost` doesn't handle required fields, so we're still stuck with unwrapping `Option` manually.

I also took the liberty of generating the structs JIT with `build.rs` rather than using a separate `codegen` step. This slows down CI for Rust significantly since we have to run everything through the Docker container, but this approach is more idiomatic and won't clutter the tree with generated files. We may want to look into caching the build container somehow.


kommendorkapten (dismissed):
Thanks!

woodruffw (dismissed):
LGTM, thank you @tnytown!

JIT generation is indeed more idiomatic, although it's slightly out-of-sync with what the rest of this repo does. I have no strong opinion here, though, so I'm happy to merge with this approach assuming @kommendorkapten and @haydentherapper have no objections.

woodruffw:
(Planning note: once merged, I'll do another RC release to crates.io.)

woodruffw:
Let's get this in following #213. That way we can do a `0.3.0` release of the crate with these new APIs.

haydentherapper:
https://github.com/sigstore/protobuf-specs/blob/1ef5b8eac3838c9ca4480ba0af494b15d5c66cee/gen/pb-rust/Cargo.toml#L3 has been updated, do we need to update under the generated code too?

haydentherapper (approved):

PR #210, opened 2024-02-01
Clarified how clients should work with the trust root during rotations

kommendorkapten:
#### Summary
Clarified how clients should interpret the trust root with respect to rotations of trusted material

#### Release Note
* Clarified how clients should interpret the trust root with respect to rotations of trusted material

#### Documentation
N/A

ping: @loosebazooka @codysoyland @bdehamer @haydentherapper @woodruffw


protos/sigstore_trustroot.proto
```diff
@@ -81,10 +81,19 @@ message CertificateAuthority {
...
 // be verified.
-// The currently used instances MUST NOT have their 'end' timestamp set in
-// their 'valid_for' attribute for easy identification.
+//
 // All the listed instances SHOULD be sorted by the 'valid_for' in ascending
-// order, that is, the oldest instance first and the current instance last.
+// order, that is, the oldest instance first. Only the last instance is
+// allowed to have their 'end' timestamp unset. All previous instances MUST
+// have a closed interval of validity. The last instance MAY have a closed
+// interval.
```
haydentherapper, line 89: > The last instance MAY have a closed interval.

What is the use case for this?

protos/sigstore_trustroot.proto
```diff
@@ -81,10 +81,19 @@ message CertificateAuthority {
...
+//
+// To be able to manage planned rotations of either transparency logs or
+// certificate authorities, clienst MUST accept lists of instances where
+// the last instance have a 'valid_for' that belongs to the future.
```
haydentherapper, line 93: +1, thanks for mentioning this.

protos/sigstore_trustroot.proto
```diff
@@ -81,10 +81,19 @@ message CertificateAuthority {
...
+// This should not be a problem as clients SHOULD first seek the trust root
+// for a suitable instance before creating a per artifact trust root (that
+// is, a sub-set of the complete trust root) that is used for verification.
```
haydentherapper, line 96: Should we mention that it's OK to have overlapping intervals of validity? For example, to handle when the rotation occurs

protos/sigstore_trustroot.proto
kommendorkapten: I can't think of a good reason _I_ wanted to do so, but as an example, the current Fulcio root expires `Not After : Oct  5 13:56:58 2031 GMT` and I think it should be ok to have that value as `end`. 

And also this makes the language clear that clients must not treat the last instance with an open interval in any specific way.

protos/sigstore_trustroot.proto
kommendorkapten: Yes, good call out, I'll update.

haydentherapper (approved):

woodruffw (approved):
LGTM!

PR #212, opened 2024-02-02
Merge `PublicKeyDetails` and `KnownSignatureAlgorithms`

kommendorkapten:
#### Summary
Merge the two messages to avoid confusion and drift.

#### Release Note
* This update deprecates the deterministic ECDSA and existing RSA algorithms in `PublicKeyDetails`. New algorithms for RSA is added, see below for details. The new algorithms for RSA specifies both the hash algorithm used, and the bit length of the public modulus.
* This update contains more algorithms in `PublicKeyDetails`
    * `PKIX_RSA_PKCS1_2048_SHA256`
    * `PKIX_RSA_PKCS1_3072_SHA256`
    * `PKIX_RSA_PKCS1_4096_SHA256`
    * `PKIX_ECDSA_P384_SHA_384`
    *  `PKIX_ECDSA_P521_SHA_512`

#### Documentation
N/A


woodruffw (approved):
LGTM, thanks @kommendorkapten! 

I'll rebase #199 on top of this.

protos/sigstore_common.proto
```diff
@@ -45,22 +45,40 @@ enum HashAlgorithm {
...
-        PKCS1_RSA_PSS = 2; // See RFC8017
-        PKIX_RSA_PKCS1V5 = 3;
-        PKIX_RSA_PSS = 4;
+        PKCS1_RSA_PKCS1V5 = 1 [deprecated = true]; // See RFC8017
+        PKCS1_RSA_PSS = 2 [deprecated = true]; // See RFC8017
+        PKIX_RSA_PKCS1V5 = 3 [deprecated = true];
+        PKIX_RSA_PSS = 4 [deprecated = true];
+        PKIX_RSA_PKCS1_2048_SHA256 = 9;
+        PKIX_RSA_PKCS1_3072_SHA256 = 10;
+        PKIX_RSA_PKCS1_4096_SHA256 = 11;
```
haydentherapper, line 71: @kommendorkapten Sorry, just reviewed - Can we keep the PSS variants?

woodruffw: Whoops, missed this as well. I can add those to #199, since I'm already modifying the file.

haydentherapper: Thanks. Do we also want to be pedantic about pkcs1 vs pkcs1v5? Also, should we support the pkcs1 encoded variants? sigstore/sigstore is able to unmarshal either pkix or pkcs1 encodings.

haydentherapper: (moving convo over to 199)

woodruffw: Hmm, I could go either way on PKCS#1 vs. PKCS#1v1.5 -- I think it's common enough in practice to abbreviate the latter as the former. But specificity also wouldn't hurt.

In terms of encodings: I would lean towards the PKIX variants, since they're what appears in X.509. But maybe the trust root encodes them as PKCS1? I'm not sure 😅

PR #214, opened 2024-02-05
protos: drop EXPERIMENTAL_ prefix

woodruffw:
Rather than mark these with `EXPERIMENTAL_` (which would force a breaking change later), this keeps the LMS and LM-OTS variant names as we expect, but with a comment emphasizing that they aren't currently supported by the PGI.


protos/sigstore_common.proto
```diff
@@ -85,9 +85,13 @@ enum PublicKeyDetails {
...
         PKIX_ED25519_PH = 8;
 
-        // **EXPERIMENTAL**: LMS and LM-OTS variants.
-        EXPERIMENTAL_LMS_SHA256 = 14;
-        EXPERIMENTAL_LMOTS_SHA256 = 15;
+        // LMS and LM-OTS
+        //
+        // These keys and signatures may be used by private Sigstore
+        // deployments, but are not currently supported by the public
+        // good instance.
```
haydentherapper, line 92: Do you think a comment along the lines of "here be dragons" would be good? This would cover some of the original motivation of the experimental prefix, that when using these, you should understand the cryptographic and performance consequences.

woodruffw: Good idea! I'll add some additional warning language here.e

kommendorkapten (approved):
Great job 🚀

PR #227, opened 2024-02-13
rust: make cargo build slightly more debuggable

woodruffw:
Handing this PR off to @tnytown, who will switch us back to AOT generation, once again completing the cycle of pain.


woodruffw:
To summarize:

* We were previously trying to release the `prost`-based crate with JIT generation via `build.rs`.
* This was causing `cargo publish` to fail, since `cargo publish` uses a different `CARGO_MANIFEST_DIR` than `cargo build`, meaning that the relative paths to the `protos` directory wasn't correct. This is seemingly completely undocumented.
* In a way, it's "good" that this was failing: if we had published this crate it would have failed even harder at `cargo install` time, since the JIT generation assumes the presence of the protos (including `protoc`'s WKTs and the Google API types).

To fix all of that, we would have had to check all of the `.proto` files into the crate as well, meaning that we get very little benefit out of JIT generation. So we're going to switch back to AOT generation and hopefully simplify everything here drastically, at the cost of slightly more checked-in code.

woodruffw:
LGTM, thanks @tnytown!

haydentherapper (approved):

gen/pb-rust/sigstore-protobuf-specs-codegen/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "sigstore-protobuf-specs-codegen"
+version = "0.0.1"
```
haydentherapper, line 3: (whoops comment didnt post) Does this version need to be in sync with the protobuf-specs version?

woodruffw: Nope, this is purely an internal, helper-only crate (it performs the codegen for the actual crate, which will remain in sync with the protobuf-specs version).

PR #228, opened 2024-02-13
Update java build workflow

haydentherapper:
This updates the generation and upload of the provenance, which broke due to the actions/artifact-upload breaking change. We now generate a draft release first, then generate provenance and upload to the existing draft release.


cpanato (approved):
we should do this for the other repos that have this , i will check

PR #237, opened 2024-02-22
Add issue templates

haydentherapper:
Since we've added an issue template in the repo, it means we won't inherit templates from the organization (https://github.com/sigstore/.github/tree/main/.github/ISSUE_TEMPLATE). This duplicates the bug and enhancement templates

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,17 @@
+---
+name: Bug
+about: Something wrong with the project? Report it here!
+labels: 'bug'
+assignees: ''
+
+---
+
+**Description**
+
+<!--Tell us about the bug. What went wrong? How can we reproduce it? What should the correct behavior be?
+-->
+
+**Version**
+
+<!--What version are you using? e.g. `cosign version`...
+-->
--- a/.github/ISSUE_TEMPLATE/issue-or-enhancement.md
+++ b/.github/ISSUE_TEMPLATE/issue-or-enhancement.md
@@ -1,0 +1,14 @@
+---
+name: Issue or Enhancement
+about: A task or enhancement idea related to the technical aspects of
+  the project.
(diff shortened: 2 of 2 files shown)


.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,17 @@
+---
+name: Bug
+about: Something wrong with the project? Report it here!
+labels: 'bug'
+assignees: ''
+
+---
+
+**Description**
```
kommendorkapten, line 9: Is this the standard template for Sigstore project?

If not I would think about having sections (beyond just description and version):
* What did you observe
* What was the expected outcome
* Steps to reproduce

haydentherapper: This is the standard template from https://github.com/sigstore/.github/tree/main/.github/ISSUE_TEMPLATE, I didn't want to change it. We could update the org-wide issue templates first?

woodruffw:
As a side-note: if we want to be even more structured here, we could use [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) instead of templates.

.github/ISSUE_TEMPLATE/bug.md
haydentherapper: FWIW, the description asks for these details.

haydentherapper:
> As a side-note: if we want to be even more structured here, we could use [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) instead of templates.

Oh neat! I'll take a look at that.

woodruffw:
> Oh neat! I'll take a look at that.

This reminded me that I meant to do it for `pip-audit` months ago, so if you want some reference samples 🙂: https://github.com/pypa/pip-audit/pull/741

PR #245, opened 2024-03-05
Clarify what is interpretted as trust anchor for chain

haydentherapper:
This standardizes client behavior on considering the last certificate in the chain as a trust anchor, regardless if it's a root or intermediate. This means clients should use all preceding CA certificates as chain-builders or intermediates.

We do document the case where an intermediate may be issued from multiple roots, specifying that multiple chains should be provided. Another solution for a later PR could be to provide two pools, trusted and untrusted, as an alternative to chains.


haydentherapper:
@woodruffw jsut to check, is it expected that `[gen/pb-rust/sigstore-protobuf-specs/src/generated/file_descriptor_set.bin](https://github.com/sigstore/protobuf-specs/pull/245/files#diff-2481ead977c11c843867edf07aac10bd147af87f4a2b709da3dfe06e715bba6d)` is updated?

kommendorkapten (approved):
Thanks!

haydentherapper:
Rebased

woodruffw:
> is it expected that `gen/pb-rust/sigstore-protobuf-specs/src/generated/file_descriptor_set.bin` is updated?

I believe so, but CCing @tnytown to confirm 🙂

woodruffw (approved):
LGTM!

PR #250, opened 2024-03-06
Check tag exists on main on release

haydentherapper:
This mitigates the risk of a tag being generated and pushed off of a branch.

### DIFF
--- a/.github/workflows/java-build-for-release.yml
+++ b/.github/workflows/java-build-for-release.yml
@@ -16,10 +16,10 @@ jobs:
       - name: Verify tag on branch
         # Check main is one of the branches where the tagged commit exists
         run: |
-          branches=$(git branch --contains ${{ github.ref }}) --format "%(refname:short)"
+          branches=($(git branch --contains ${{ github.ref }} --format "%(refname:short)"))
           found=false
-          for b in $raw; do
-            if [ "$x" = "main" ]; then
+          for b in "${branches[@]}"; do
+            if [ "$b" = "main" ]; then
               found=true
             fi
           done
--- a/.github/workflows/java-build-for-release.yml
+++ b/.github/workflows/java-build-for-release.yml
@@ -5,10 +5,34 @@ on:
       # if you change this pattern, make sure jobs.strip-tag still works
       - 'release/java/v[0-9]+.[0-9]+.[0-9]+'
 jobs:
+  check-tag-main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout all branches
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          fetch-depth: 0
+
+      - name: Verify tag on branch
+        # Check main is one of the branches where the tagged commit exists
+        run: |
+          branches=$(git branch --contains ${{ github.ref }}) --format "%(refname:short)"
+          found=false
+          for b in $raw; do
+            if [ "$x" = "main" ]; then
+              found=true
+            fi
+          done
+          if [ "$found" = false ]; then
+            exit 1
+          fi
+
   ci:
+    needs: [check-tag-main]
     uses: ./.github/workflows/java-build.yml
 
   strip-tag:
+    needs: [check-tag-main]
     runs-on: ubuntu-latest
     outputs:
       version: ${{ steps.version.outputs.version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/java-build-for-release.yml
```diff
@@ -5,10 +5,34 @@ on:
...
+        with:
+          fetch-depth: 0
+
+      - name: Verify tag on branch
+        # Check main is one of the branches where the tagged commit exists
+        run: |
+          branches=$(git branch --contains ${{ github.ref }}) --format "%(refname:short)"
+          found=false
+          for b in $raw; do
+            if [ "$x" = "main" ]; then
```
woodruffw, line 22: Wrong variable, perhaps? 🙂 

```suggestion
            if [ "$b" = "main" ]; then
```

.github/workflows/java-build-for-release.yml
```diff
@@ -5,10 +5,34 @@ on:
...
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          fetch-depth: 0
+
+      - name: Verify tag on branch
+        # Check main is one of the branches where the tagged commit exists
+        run: |
+          branches=$(git branch --contains ${{ github.ref }}) --format "%(refname:short)"
+          found=false
+          for b in $raw; do
```
woodruffw, line 21: Slight nitpick: you can expand into an array here and then use a bit of dedicated syntax to expand into the loop more idiomatically (IMO):

```suggestion
          branches=($(git branch --contains ${{ github.ref }}) --format "%(refname:short)")
          found=false
          for b in "${branches[@]}"; do
```

woodruffw:
Thinking out loud: 

1. I _think_ this will be unreliable in a future release scenario where `main` tracks the latest stable release, but we have "backport" branches that get cherry-picked into. For example, `main` might track the `2.x` series, while the `1.2.3` tag gets created off of the `1.x` branch by cherry-picking from `main`. We may never run into this as a practical matter, but I figured it's worth flagging as something that could break in the future 🙂 
2. This might not be a sufficient protection in all cases: in particular, someone with tag push access might still be able to push up a tag that has a modified version of this workflow, removing this check. I'm not 100% sure about that though, someone from GitHub like @steiza might be able to confirm.

(This would be a bigger change, but I think we could solve both of these problems by switching from tag events to `release: published` events -- that event payload includes the `GITHUB_REF` so we can differentiate for each language, while also giving us more control over who can actually publish the packages via release signoffs.)

haydentherapper:
> I think this will be unreliable in a future release scenario where main tracks the latest stable release, but we have "backport" branches that get cherry-picked into. For example, main might track the 2.x series, while the 1.2.3 tag gets created off of the 1.x branch by cherry-picking from main. We may never run into this as a practical matter, but I figured it's worth flagging as something that could break in the future 🙂

Yea, to generalize this check, it's really "am i pushing from a protected branch", assuming we create X.Y branches that are protected for previous releases. Sticking with main for now is probably sufficient, unless there's a straightforward way to know if the branch is protected.

> This might not be a sufficient protection in all cases: in particular, someone with tag push access might still be able to push up a tag that has a modified version of this workflow, removing this check. I'm not 100% sure about that though, someone from GitHub like @steiza might be able to confirm.

Ugh, good point, I think you're right that the action would run based on what's in the tagged release, invalidating this check.

Would switching to `release:published` help with preventing pushes from non-protected branches? Couldn't you still publish a release from a tag that is from a branch?

haydentherapper:
Adding a few others for thoughts: @kommendorkapten @loosebazooka

haydentherapper:
The other option is to simply trigger releases manually from what's in main.

woodruffw:
> Would switching to `release:published` help with preventing pushes from non-protected branches? Couldn't you still publish a release from a tag that is from a branch?

I think you could, assuming that user has release creation permissions. However, the release workflow itself could be put in an environment with a configured "required reviewers" rule, to prevent someone from unilaterally publishing a package through a release.

For example, here's what our `release` environment looks like on `pip-audit`:
Screenshot 2024-03-06 at 6 11 44 PM

(and in the case of `protobuf-specs`, I figure we'd want the "Prevent self-review" option as well 🙂)

kommendorkapten:
Yay to using environments (and prevent self-review) as it's a good way to force manual approval before a workflow is run. We don't release that often, and adding a a manual control is worth it I believe. 

Making the switch from tags to release events seems solid. I'm thinking if we should keep workflows for tags for alfa, beta or pre-releases though, as it can be good testing without requiring extra assistance.

haydentherapper:
https://github.com/sigstore/protobuf-specs/issues/254 to track, closing this out.

Issue #244, opened 2024-03-05
Permissions error running make locally


haydentherapper:
There is a permissions error when trying to read the lock generated during rust codegen. This appears to surface after the rust codegen is complete and a later step accesses the lock (which appears to occur during json schema generation)

```
Building docker image for generating JSON schema files
docker build -t jsonschema-specs-build -f Dockerfile.jsonschema .
error checking context: no permission to read from '/usr/local/google/home/hayden/protobuf-specs/gen/pb-rust/target/debug/incremental/sigstore_protobuf_specs_codegen-233wpt7eagx8m/s-gu0xx5y87x-120l4cw.lock'
make: *** [Makefile:89: docker-image-jsonschema] Error 1
```

Running with `sudo` fixes this, but obviously is not ideal (and seems to create other errors for me, but that's probably due to my environment). 

cc @woodruffw @tnytown

haydentherapper:
Oh this might be more than just Rust, I see when I run `make clean`, I get permission denied errors for all generated files. I assume this is because the docker container generated the files? I guess we need to chown after generation?

kommendorkapten:
I think we should avoid running docker as root, let me see if I can take a quick look at this.

kommendorkapten:
@haydentherapper please take a look at https://github.com/sigstore/protobuf-specs/pull/251 and see if it works better (you first need to clean up the existing files of course)

bobcallaway:
fixed in #473

PR #251, opened 2024-03-07
Don't run as root when building protobuf files

kommendorkapten:
May close https://github.com/sigstore/protobuf-specs/issues/244

#### Summary
Make sure that docker does not run as root when building protobuf files.

#### Release Note
N/A

#### Documentation
N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,15 +6,23 @@ RUN set -ex && \
     apt-get install -y --no-install-recommends \
     python3-pip
 
-# Install Rust cargo.
-RUN set -ex && \
-    apt-get install -y --no-install-recommends \
-        curl \
-        build-essential
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- -y
-ENV PATH="/root/.cargo/bin:${PATH}"
-
 # Install Python dev dependencies.
 COPY ./dev-requirements.txt /tmp/
 RUN python3 -m pip install --upgrade pip && \
     python3 -m pip install --requirement /tmp/dev-requirements.txt
+
+# Install Rust cargo.
+RUN set -ex && \
+    apt-get install -y --no-install-recommends \
+        curl \
+    build-essential
+
+# Switch user
+ARG uid=1000
+RUN useradd -u ${uid} -s /bin/sh -m builder
+
+USER builder
+WORKDIR /home/builder
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- -y
+ENV PATH="/home/builder/.cargo/bin:${PATH}"
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN python3 -m pip install --upgrade pip && \
 RUN set -ex && \
     apt-get install -y --no-install-recommends \
         curl \
-    build-essential
+        build-essential
 
 # Switch user
 ARG uid=1000
--- a/Dockerfile.jsonschema
+++ b/Dockerfile.jsonschema
@@ -1,7 +1,19 @@
 # 3.18.2
 FROM alpine@sha256:c5b1261d6d3e43071626931fc004f70149baeba2c8ec672bd4f27761f8e1ad6b
 RUN apk add --update protoc protobuf-dev go git
+
+# Switch user
+ARG uid=1000
+RUN adduser -u ${uid} -S builder
+
+RUN mkdir -p /home/builder
+RUN chown builder /home/builder
+
+USER builder
+WORKDIR /home/builder
+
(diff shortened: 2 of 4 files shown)


kommendorkapten:
This is a quite massive rewrite as getting everything to run as a non-root user isn't trivial. I would appreciate if as many as possible could run 
```shell
$ make clean
$ make
$ git status
```

And verify that it all builds correctly and no un-expected modifications happens to the generate files.

Dockerfile.jsonschema
```diff
@@ -1,7 +1,19 @@
 # 3.18.2
 FROM alpine@sha256:c5b1261d6d3e43071626931fc004f70149baeba2c8ec672bd4f27761f8e1ad6b
 RUN apk add --update protoc protobuf-dev go git
+
+# Switch user
+ARG uid=1000
+RUN adduser -u ${uid} -S builder
```
kommendorkapten, line 7: Alpine does not use the "standard" `useradd` command.

Makefile
```diff
@@ -21,72 +21,69 @@ RUST_ACTION ?= run -p sigstore-protobuf-specs-codegen
...
 		--entrypoint bash ${PROTOC_IMAGE} \
-		-c "cd gen/pb-rust && cargo ${RUST_ACTION}"
+		-c "cd /defs/gen/pb-rust && cargo ${RUST_ACTION}"
 
 # docker already does its own caching so we can attempt a build every time
 .PHONY: docker-image
 docker-image:
 	@echo "Building development docker image"
-	docker build -t ${PROTOC_IMAGE} .
+	${DOCKER_BUILD} -t ${PROTOC_IMAGE} .
```
loosebazooka, line 73: If you pass in the current host system user and group as args here you can use those IDs when creating the user and group in the docker file instead of 1000

kommendorkapten: That's actually what I'm doing. They are passed as build-args, and if provided they are used instead of the default value of 1000.

kommendorkapten: Or only the user id to be exact. But it should be enough.

Makefile
```diff
@@ -21,72 +21,69 @@ RUST_ACTION ?= run -p sigstore-protobuf-specs-codegen
 # generate all language protobuf code
 all: go python typescript ruby jsonschema rust
 
+CUSER=$(shell id -u ${USER})
+CGROUP=$(shell id -g ${USER})
+DOCKER_RUN=docker run --platform linux/amd64 -u ${CUSER}:${CGROUP} -v ${PWD}:/defs
```
kommendorkapten, line 26: Here and the line below is where the current system user-id is passed @loosebazooka

loosebazooka:
Nice! Works for me.

Dockerfile
```diff
@@ -6,15 +6,23 @@ RUN set -ex && \
     apt-get install -y --no-install-recommends \
     python3-pip
 
+# Install Python dev dependencies.
```
haydentherapper, line 9: Testing locally, I had to move lines 14-18, the apt-get for curl and build-essential, above 10-12, or else it failed with `E: Failed to fetch http://deb.debian.org/debian/pool/main/p/perl/perl-modules-5.32_5.32.1-4%2bdeb11u2_all.deb  404  Not Found [IP: 151.101.22.132 80]` (not sure why). Can we swap the order?

Dockerfile
```diff
@@ -6,15 +6,23 @@ RUN set -ex && \
...
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- -y
-ENV PATH="/root/.cargo/bin:${PATH}"
 
-# Install Python dev dependencies.
-COPY ./dev-requirements.txt /tmp/
-RUN python3 -m pip install --upgrade pip && \
-    python3 -m pip install --requirement /tmp/dev-requirements.txt
+# Switch user
+ARG uid=1000
+RUN useradd -u ${uid} -s /bin/sh -m builder
```
haydentherapper, line 22: I got an error: `adduser: number 561799 is not in 0..256000 range`. Looks like https://stackoverflow.com/questions/41807026/cant-add-a-user-with-a-high-uid-in-docker-alpine would be the fix?

haydentherapper (changes requested):

haydentherapper:
This is awesome, thanks so much for tackling this.

Dockerfile
kommendorkapten: Interesting, I'll try to take a look later this week.

haydentherapper:
@kommendorkapten, did you get a chance to look into the open comments, or would you like someone to take a look?

kommendorkapten:
@haydentherapper Last weeks was busy with preparation for KubeCon and so. I'm back now from the Easter break so I would be able to work on this again this week.

kommendorkapten:
Closing in favour of https://github.com/sigstore/protobuf-specs/pull/473

PR #262, opened 2024-03-19
Add Rekor Bundle message

### LINKED ISSUE #261: Add canonicalized RekorBundle to the Rekor proto
Based on the discussion about [verifying the SET](https://github.com/sigstore/rekor/issues/1943#issuecomment-1968352634), I'm opening this issue to add a canonicalized RekorBundle to the TLE, to be used by Rekor in its CLI output. The purpose is to have a way to easily get the SET's corresponding payload from the `rekor-cli` in order to manually and auditably verify an SET, rather than having to construct the payload from bits and pieces of output, which is very cumbersome.

I've had some confusion around the "Sigstore bundle" vs. the "Rekor bundle" so restating my understanding: It looks like the Sigstore bundle `VerificationMaterial` has an [embedded `TransparencyLogEntry`](https://github.com/sigstore/protobuf-specs/blob/a57a7caac5f4daf428055a34c872c810a6c0a968/protos/sigstore_bundle.proto#L105), inside which would be the place for the Rekor bundle. I know the Rekor bundle is what is signed by the Rekor public key to produce the SET, which is part of the verification material.

Possibly related issues:
https://github.com/sigstore/protobuf-specs/issues/11 
https://github.com/sigstore/protobuf-specs/issues/116

lkatalin:
Related to #261 
Related to https://github.com/sigstore/rekor/issues/1943 

I'm considering whether the RekorBundle struct would make more sense as bytes (like the `canonicalized body`), with the tradeoff of readability, so I may update this. (Or possibly we could have both, but that may be too verbose.)

Edit: Updated based on the conversation in #261 to be separate from the TLE message.


protos/sigstore_rekor.proto
```diff
@@ -128,3 +128,12 @@ message TransparencyLogEntry {
         // payload from other sources to verify the signature.
         bytes canonicalized_body = 7;
 }
+
+// The RekorBundle is the signed material used to produce the Signed Entry
+// Timestamp signature. See notes on the InclusionPromise above.
+message RekorBundle {
```
haydentherapper, line 134: Have we confirmed that when marshalled as json as canonicalized, this works as expected? Just want to make sure there's no issues with field naming.

lkatalin: I can run some tests soon. No need to wait on this PR for the new release.

woodruffw, line 134: Layout LGTM here, but I'm kind of murky on where this structure ends up being used (since it isn't referenced in the other messages). Is the idea that a tool like `rekor-cli` is expected to use it?

haydentherapper: I believe the idea is it could replace structure declarations like https://github.com/sigstore/sigstore-go/blob/main/pkg/tlog/entry.go#L260-L265, so that clients don't have to know the contents of the SET. They would still need to know the mapping between the TLE structure and this one though.

haydentherapper, line 134: naming nit: I don't want to proliferate "bundle" even though it's in cosign, given we want "bundle" to refer to top level protobuf bundle now. Do you have another suggestion of the name? SignedEntryTimestampContents?

woodruffw: Gotcha, makes sense!

haydentherapper:
Going to close this out, as Rekor v2 supersedes this and reuses the `TransparencyLogEntry`.

PR #277, opened 2024-03-26
trustroot: initial client config messages

### LINKED ISSUE #259: Add infrastructure information to trusted root
For a given trusted root (especially with sigstore public good), the fulio/ca instance should optionally be allowed to specify a(or many?) sigstore specific trusted oidc provider(s).

Why?
This allows clients to fully initialize their state from a single trusted_root.json, currently the only information missing is what would be required to initialize the oidc client for dex.

For example:
```javascript
  "certificateAuthorities": [
    {
      "subject": {
        "organization": "sigstore.dev",
        "commonName": "sigstore"
      },
      "uri": "https://fulcio.sigstore.dev",
      //... new oidc providers associated with an instance of fulcio
      oidcProviders: [
        {
           "uri": "oauth2.sigstore.dev/auth"
           "clientId": "sigstore" // should default to this
           "wellKnownConfiguration": "/.well-known/openid-configuration"
        }
      ]
    },
```
1. This should be optional?
    - Yes? some instances may not want a web oidc option. Also allows us to clear the config is archived CAs.
1. should it be a list?
    - Yes? Allows for extensibility or single config for internal/external usecases
    - No? Client logic now has to check for existence/availability of endpoints as it decides which oidc endpoint to contact
1. should it be per CA?
    - Yes? It's not particularly harder for clients to parse this information.
    - Yes? As the trusted_root is updated, old CAs that are no longer *active*, will not need to define this. So repetition will be minimal.
1. is URI enough or should we have a bunch of other fields like `client_id`, `well_known_configuration`
    - Yes? Extensibility, custom configs

woodruffw:
~~WIP; needs workshopping 🙂.~~ Closes #259.

CC @haydentherapper @loosebazooka @kommendorkapten


protos/sigstore_trustroot.proto
```diff
@@ -117,3 +117,43 @@ message TrustedRoot {
...
+        // A URL to a Fulcio-compatible CA, capable of receiving
+        // Certificate Signing Requests (CSRs) and responding with
+        // issued certificates.
+        //
+        // This URL **MUST** be the "base" URL for the CA, which clients
+        // should construct an appropriate CSR endpoint on top of.
+        // For example, if `fulcio_url` is `https://example.com/ca`, then
+        // the client **MAY** construct the CSR endpoint as
+        // `https://example.com/ca/api/v2/signingCert`.
+        string fulcio_url = 1;
```
woodruffw, line 135: I was originally going to call this `csr_url` (and similar `tlog_url` for Rekor below), but in practice the APIs here are Fulcio- and Rekor-specific. Curious if anybody has strong thoughts here.

protos/sigstore_trustroot.proto
```diff
@@ -117,3 +117,43 @@ message TrustedRoot {
...
+        //
+        // This URL **MUST** be the "base" URL for the OIDC IdP, which clients
+        // should perform well-known OpenID Connect discovery against.
+        string oidc_url = 2;
+
+        // A URL to a Rekor-compatible transparency log.
+        //
+        // This URL **MUST** be the "base" URL for the transparency log,
+        // which clients should construct appropriate API endpoints on top of.
+        string rekor_url = 3;
```
loosebazooka, line 147: do we need a timestamp authority url?

protos/sigstore_trustroot.proto
loosebazooka: I think it would be more consistent if they matched the types above?

protos/sigstore_trustroot.proto
woodruffw: Ah, yep, good catch!

sigstore-python doesn't currently do any TSA stuff, so to confirm: would this URL be expected to be the base endpoint for receiving TSRs, or should the client know to append/build an endpoint on top of it?

woodruffw: Hmm, I don't think that's standard 😅 -- I believe RFC 3161 only stipulates an HTTP POST to a pre-defined endpoint of a DER-encoded TSR. I'll look into it.

protos/sigstore_trustroot.proto
woodruffw: I agree, although that _might_ be confusing: technically verification requires none of the machinery specific to Rekor or Fulcio (so we can speak in general CA, CT, etc. terms), while signing _concretely_ relies on REST APIs exposed by Fulcio or Rekor (or services imitating those).

Then again, maybe that can be solved by just documenting in the comments that these URLs are expected to adhere to the Fulcio/Rekor APIs. If that seems reasonable to you, I'll rename and update for that 🙂

protos/sigstore_trustroot.proto
woodruffw: Yeah, RFC 3161 is pretty badly underspecified here: it says HTTP, but doesn't specify an HTTP method or a particular endpoint. It _does_ however stipulate a content-type and that the payload is just the bare DER for the TSR, which isn't true in this case. So I _think_ we want to stipulate that the URL here behaves like <https://github.com/sigstore/timestamp-authority>.

CC @haydentherapper for opinions here 🙂

loosebazooka: Yeah, I looked up other TSAs and it's all a bit different. I don't know if @bdehamer has an opinion. Like should the TSA endpoint just be a full address?

protos/sigstore_trustroot.proto
loosebazooka: Yeah seems reasonable.

protos/sigstore_trustroot.proto
woodruffw: Yeah, I think we should probably just stipulate that this is a Sigstore-shaped TSA. There isn't enough normative standards language to really enforce anything here and it only matters for signing (since verifying is universal across TSAs).

loosebazooka: Would any rfc3161 timestamp satisfy the requirements of sigstore?

protos/sigstore_trustroot.proto
woodruffw: PTAL!

protos/sigstore_trustroot.proto
woodruffw: Added `tsa_url` with docs explaining that it should behave like the Sigstore TSA.

woodruffw: > Would any rfc3161 timestamp satisfy the requirements of sigstore?

On the verifying side yeah, but on the signing side we need to know *how* to submit the TSR to the TSA. That part isn't well-specified, hence the need for concrete behavior here 🙂

haydentherapper: Sorry if I was unclear, I think we should do the opposite and require the full path. The RFC specifies the expected payload and content type. You're right that it doesn't specify HTTP method, but in practice, all of the TSAs we tested used POST. The [golang implementation](https://github.com/sigstore/cosign/blob/main/internal/pkg/cosign/tsa/client/client.go#L120 ), with the RFC-compliant TSR and content header, worked for more than just the sigstore TSA as long as the full API path was provided.

woodruffw: Nope, I just somehow missed your message 😅 -- I'm going to blame that on conference Wi-Fi. 

I totally missed that the Sigstore TSA accepts the `application/timestamp-query` content-type in addition to JSON (which is what sigstore-js seems to use). Given that, fully agreed on passing the full URL, and I'll update for that 🙂

woodruffw: Updated, PTAL!

haydentherapper: Ah yes, we added JSON later on (https://github.com/sigstore/timestamp-authority/pull/247) for clients that didn't want to deal with DER, but the TSA impl is fully RFC3161 compliant.

protos/sigstore_trustroot.proto
```diff
@@ -117,3 +118,50 @@ message TrustedRoot {
...
+        // A URL to a Fulcio-compatible CA, capable of receiving
+        // Certificate Signing Requests (CSRs) and responding with
+        // issued certificates.
+        //
+        // This URL **MUST** be the "base" URL for the CA, which clients
+        // should construct an appropriate CSR endpoint on top of.
+        // For example, if `ca_url` is `https://example.com/ca`, then
+        // the client **MAY** construct the CSR endpoint as
+        // `https://example.com/ca/api/v2/signingCert`.
+        string ca_url = 1;
```
haydentherapper, line 136: Aligning with the above names seems better for readability, so +1 with what's here.

I did bring up a related [issue](https://github.com/sigstore/sigstore-go/issues/71) in sigstore-go where I suggested we call tlog related functions `RekorLogs` as Rekor is the only valid log. I think the same argument could be applied to Fulcio given we couldn't support a generic CA without standardization on request payload and headers. I could see an argument to calling these `(rekor|fulcio)_url`, though we ideally would also change the above to `rekor_logs`/`fulcio_cas`. Given I don't think we should do a breaking change just for names, I'd stick with `(ca|tlog)_url`.

protos/sigstore_trustroot.proto
```diff
@@ -117,3 +118,50 @@ message TrustedRoot {
...
+        // This URL **MUST** be the **full** URL for the TSA, meaning that it
+        // should be suitable for submitting Time Stamp Requests (TSRs) to
+        // via HTTP, per RFC 3161.
+        string tsa_url = 4;
+}
+
+// ClientTrustConfig describes the complete state needed by a client
+// to perform both signing and verification operations against a particular
+// instance of Sigstore.
+message ClientTrustConfig {
```
haydentherapper, line 161: :chefs-kiss on message naming

protos/sigstore_trustroot.proto
```diff
@@ -117,3 +118,50 @@ message TrustedRoot {
...
+        //
+        // This URL **MUST** be the "base" URL for the OIDC IdP, which clients
+        // should perform well-known OpenID Connect discovery against.
+        string oidc_url = 2;
+
+        // A URL to a Rekor-compatible transparency log.
+        //
+        // This URL **MUST** be the "base" URL for the transparency log,
+        // which clients should construct appropriate API endpoints on top of.
+        string tlog_url = 3;
```
haydentherapper, line 148: Should we support multiple URLs for multiple log entries and timestamps? [`VerificationMaterial`](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto#L59) supports this already. Hopefully down the road, we'll see additional logs spun up, similar to the Certificate Transparency ecosystem, where users can submit entries to multiple logs as a mitigation against an individual log presenting a split-view.

We could say YAGNI, but it might be easier to do this now rather than have to have a breaking change later. And it shouldn't be much work for clients who ingest this message to request timestamps and log proofs for each URL, adding each to the respective repeated bundle field.

woodruffw: Yeah, that makes sense to me! I'll make those changes.

woodruffw: [1478df1](https://github.com/sigstore/protobuf-specs/pull/277/commits/1478df19e31c3c26967ef8c549c475e7b5fe7301)

protos/sigstore_trustroot.proto
woodruffw: Thanks!

protos/sigstore_trustroot.proto
```diff
@@ -117,3 +118,50 @@ message TrustedRoot {
         // A set of trusted timestamping authorities.
         repeated CertificateAuthority timestamp_authorities = 5;
 }
+
+
+// SigningConfig represents the trusted entities/state needed by Sigstore
+// signing. In particular, it primarily contains service URLs that a Sigstore
+// signer may need to connect to for the online aspects of signing.
+message SigningConfig {
```
woodruffw, line 126: Flagging: which, if any, of these should be required? I think `ca_url` and `oidc_url` at minimum.

loosebazooka: I could see a non-keyless config existing? Do we care about those people? Starting everything optional seems safest? There's a bit of ambiguity of what a proper "sigstore" signature means -- cause cosign allows everything.

woodruffw: Yeah, good point. The requirements are purely hints at the DSL level here anyways, so we can leave them out for now and expect clients to enforce their own requirements here.

loosebazooka: sgtm

protos/sigstore_trustroot.proto
```diff
@@ -117,3 +118,50 @@ message TrustedRoot {
...
+        // via HTTP, per RFC 3161.
+        repeated string tsa_urls = 4;
+}
+
+// ClientTrustConfig describes the complete state needed by a client
+// to perform both signing and verification operations against a particular
+// instance of Sigstore.
+message ClientTrustConfig {
+        // The root of trust, which MUST be present.
+        TrustedRoot trusted_root = 1 [(google.api.field_behavior) = REQUIRED];
```
kommendorkapten, line 163:
```suggestion
        TrustedRoot trusted_root = 2 [(google.api.field_behavior) = REQUIRED];
```

protos/sigstore_trustroot.proto
```diff
@@ -117,3 +118,50 @@ message TrustedRoot {
...
+
+        // Configuration for signing clients, which MUST be present.
+        SigningConfig signing_config = 2 [(google.api.field_behavior) = REQUIRED];
```
kommendorkapten, line 166:
```suggestion
        SigningConfig signing_config = 3 [(google.api.field_behavior) = REQUIRED];
```

protos/sigstore_trustroot.proto
```diff
@@ -117,3 +118,50 @@ message TrustedRoot {
...
+        // should be suitable for submitting Time Stamp Requests (TSRs) to
+        // via HTTP, per RFC 3161.
+        repeated string tsa_urls = 4;
+}
+
+// ClientTrustConfig describes the complete state needed by a client
+// to perform both signing and verification operations against a particular
+// instance of Sigstore.
+message ClientTrustConfig {
+        // The root of trust, which MUST be present.
```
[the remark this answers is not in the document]
kommendorkapten: I added a media type that follows the update in this PR https://github.com/sigstore/protobuf-specs/pull/279

I think we should wait with this PR until the media type is updated to not confuse clients too much.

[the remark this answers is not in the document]
woodruffw: SGTM!

haydentherapper:
LGTM, I think this just needs a regeneration

woodruffw:
> LGTM, I think this just needs a regeneration

Done!

haydentherapper:
@woodruffw Do you want to get this in before 0.3.1?

kommendorkapten (approved):
Thanks!

Before merging, I would just like to make sure we are agreeing on how this will be distributed via PGI TUF root (i.e how this should be documented in the Sigstore Client Spec).

We now have:
* SigningConfig
* TrustedRoot

And then ClientTrustConfig = {SigningConfig, TrustedRoot}

The TUF repository already ships TrustedRoot as single artifact, `trusted_root.json`.
How would we ship the SigningConfig? My recommendation is to add a new artifact to the TUF root called `signing_config.json`. Why? Do not break the clients. Adding a new target is trivial. 

The alternative is to add a new target `client_trust_config` that contains both messages. The downside is that we would need to duplicate the `trusted_root`. 

Shipping two files from the TUF repo is not the end of the world IMHO. We can still allow clients to accept a single file (`client_trust_config`) to simplify the UX when e.g. invoked via the terminal.

As we are in the process of adding support for trusted root to a lot of components, withdrawing that from the TUF repository would be very disruptive 😅

woodruffw:
This might be a minority opinion, but IMO it's okay for the PGI TUF repo to **not** ship `SigningConfig` at all -- the PGI and staging instances both have "well-known" signing configurations already, so a client can (continue to) hardcode them. 

In other words, IMO it's okay if `SigningConfig` and `ClientTrustConfig` exist entirely as message definitions for checking user (i.e. non-TUF) JSON inputs against, with the idea being that they establish the interface for clients that want to support CLIs like `--client-trust-config`. 

Regardless of the above, I 100% agree about not removing the current `trusted_root.json` target  -- IMO if we *want* to put `SigningConfig` or `ClientTrustConfig` in the TUF repo, then a separate file (either just the signing config or the combined client trust) would be perfectly cromulent 🙂

woodruffw:
Merging!

haydentherapper:
@woodruffw @kommendorkapten I think a primary motivation would be if we move Rekor to yearly sharding, since the URL would change periodically and clients would need to discover it without breaking old clients. I don't think we need to immediately ship a `SigningConfig` though since this isn't the case currently, but I do think we should start making plans for adding an additional file distributed through TUF.

kommendorkapten:
@haydentherapper yes, I proposed in the client meeting to today that we add the `signing_config.json` to the TUF repo during next signing.

Issue #165, opened 2023-11-07
Bundle Media Type


bdehamer:
Currently there are two valid values for the bundle `media_type` field:

* `application/vnd.dev.sigstore.bundle+json;version=0.1`
* `application/vnd.dev.sigstore.bundle+json;version=0.2`

I'm working on a project where I want to push a Sigstore bundle to an OCI registry alongside the related image.  As part of this I want to create an [OCI manifest](https://github.com/opencontainers/image-spec/blob/main/manifest.md) similar to the following:

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "artifactType": "application/vnd.dev.sigstore.bundle+json;version=0.2",
  "config": {
    "mediaType": "application/vnd.oci.empty.v1+json",
    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
    "size": 2
  },
  "layers": [
    {
      "mediaType": "application/vnd.dev.sigstore.bundle+json;version=0.2",
      "digest": "sha256:4bd9df17d3cfa8632690f6251b7dc6d2f7cebd60313c49bea4092b9489e2d4a4",
      "size": 4967
    }
  ],
  "subject": {
    "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
    "digest": "sha256:010511b82573da0735bbbc09ab0b1b9e9218732306d96b81beb694cfe431a499",
    "size": 523
  },
}
```

Note that I'm including the Sigstore media type in the manifest's `artifactType` field and as the `mediaType` for the layer which points at specific blob which contains the actual bundle.

I've found that at least one of the major image registries (AWS ECR) chokes on this manifest and it appears to be specifically related to the bundle media type. It seems to reject any value which does not strictly conform to the [RFC 6838 spec](https://datatracker.ietf.org/doc/html/rfc6838#section-4.2) (which is the one which is called-out in the [OCI image spec](https://github.com/opencontainers/image-spec/blob/main/manifest.md#image-manifest-property-descriptions)). RFC 6838 is more constrained than [RFC 2045](https://datatracker.ietf.org/doc/html/rfc2045) and does NOT allow for the sort of "version=0.2" parameter addition we're using for the Bundle media type.

Sure enough, if I remove the ";version=0.2" bit, AWS will happily accept my manifest!

[...]

PR #279, opened 2024-03-27
Updated media type to a format compatible with OCI registries

kommendorkapten:
**Note** that i did _not_ opt for bumping the version in the media type, only changing the encoding. I would be very interesting to hearing the pros/cons. As there are no semantical changes, I kept the semantic versioning. But I'm struggling to come to a resolution if that was a good or bad move. If we bump the semantic versions, some parts of the code will be more complicated client side. I'm easy to convince the semantic version should be bumped if there are more concrete use-cases where it would be simpler 😅  

#### Summary
Closes https://github.com/sigstore/protobuf-specs/issues/165

#### Release Note
Added a new format for the media type that is compatible with OCI registries

#### Documentation
N/A


woodruffw:
LGTM overall! 

I'm also divided on bumping the version here:

* On one hand, *not* bumping here will result in bundles/trust roots that clients *will* reject if (like sigstore-python) they perform strict matching on the media type.
* On the other, this isn't really a breaking change in the underlying formats themselves, and bumping the version contributes to version fatigue here.

kommendorkapten:
> On the other, this isn't really a breaking change in the underlying formats themselves, and bumping the version contributes to version fatigue here.

This is exactly the one I tried to avoid :) 

But yes, agree with all you said @woodruffw

bdehamer (approved):
I agree. Leaving the version at 0.3 feels like the right thing to do here. 

When we publish the new protobuf packages, I assume we'll treat this as a patch update to the v0.3.x family, yeah?

haydentherapper:
Are there any clients generating v0.3 bundles yet?

bdehamer:
In sigstore-js, we can verify a v0.3 bundle, but don't yet generate them.

haydentherapper (approved):
LGTM. Only question is if we have any v3 bundles floating about. If so, then this LGTM. If not, then we could change the v3 string without having to support the previous version.

haydentherapper:
sigstore-go isn't generating v3 bundles.

kommendorkapten:
What about java and python, do you produce v0.3 bundles yet? @woodruffw @loosebazooka

loosebazooka:
I think it's okay to keep it at v3 for java. It's committed in code, but unused.

woodruffw:
> I think it's okay to keep it at v3 for java. It's committed in code, but unused.

Ah yeah, I forgot that nobody has really rolled out v3 bundle support yet. Given that, I think we're safe to not bump the version here!

haydentherapper:
Do we need to support `application/vnd.dev.sigstore.bundle+json;version=0.3`, or could we remove that string so we don't have two ways of representing v3 bundles?

kommendorkapten:
> Do we need to support application/vnd.dev.sigstore.bundle+json;version=0.3, or could we remove that string so we don't have two ways of representing v3 bundles? 

That is a very tempting idea!

We do have a release/tag `v0.3.0` out which specifies the `version=0.3`. So if we remove the old version 0.3, should we remove that release? And is it used? If we remove it It would be a breaking change? So I'm leaning towards keeping it.

The language we have now is:
```
        // MUST be application/vnd.dev.sigstore.bundle.v0.3+json when
        // when encoded as JSON.
        // Clients must to be able to accept media type using the previously
        // defined formats:
        // * application/vnd.dev.sigstore.bundle+json;version=0.1
        // * application/vnd.dev.sigstore.bundle+json;version=0.2
        // * application/vnd.dev.sigstore.bundle+json;version=0.3
```

To minimize the risk of breaking anything I think we should keep as is? It's unlikely, but someone can be using the v3 version in some private scenario and I would hate to make their lives harder.

woodruffw:
Removing `version=0.3` is temping, but I don't think we should muddle with the version history 😅 

IMO, we can do the following

1. Tweak the language to say "v3 bundles MUST use the new version format, but clients MAY allow the old format for compatibility"
2. Release 0.3.1 with the changes

In practice, this should have no compatibility concerns, since no clients are generating v3 bundles at the moment.

kommendorkapten:
I think we are more or less saying the same @woodruffw, but I think the language we have is enough? But yes, `v0.3.1` tag is what I would propose too (but didn't write that) 😄

woodruffw:
> I think we are more or less saying the same @woodruffw, but I think the language we have is enough? But yes, `v0.3.1` tag is what I would propose too (but didn't write that) 😄

Yes, I think so!

haydentherapper:
SGTM, agreed we shouldn't yank a release.

woodruffw (approved):

haydentherapper:
Merging since there's 3 LGTMs

PR #284, opened 2024-04-04
Add checkpoint key ID to trust root

haydentherapper:
This adds a string to represent the checkpoint key ID for a log, which will differ for ed25519 logs. To simplify client implementation, we will provide this string so that clients don't have to compute the checkpoint key ID themselves using the public key. If it's not set, then a client should assume the log ID is equal to the checkpoint key ID, which is true for ecdsa and rsa logs.

Also cleaned up the text for the checkpoint.

Ref: https://github.com/sigstore/rekor/issues/2062


protos/sigstore_common.proto
```diff
@@ -140,6 +140,20 @@ message LogId {
...
+// For RSA signatures, the key ID will match the ECDSA format of the hashed
+// DER-encoded SPKI public key. Publicly witnessed logs MUST NOT use
+// RSA-signed checkpoints, since witnesses do not support RSA signatures.
+// This is provided for convenience. Clients can also calculate the checkpoint
+// key ID given the log's public key.
+message CheckpointKeyId {
+        // The key ID in a checkpoint, as a prefix to the signature. SHOULD be
+        // 4 bytes long, as a truncated hash.
+        bytes key_id = 1 [(google.api.field_behavior) = REQUIRED];
+}
```
woodruffw, line 155: This is a minor thing, but this will mark at least the third overlapping type for this in the protobuf-specs 😅 

* `common.LogId`
* `common.PublicKeyIdentifier`
* (new) `common.CheckpointKeyId`

I have no objection to a new type in principle, but would either of the pre-existing ones work? `PublicKeyIdentifier` in particular might be ideal, since it isn't prescriptive in name like `LogId` and has an arbitrary `string hint`.

haydentherapper: `PublicKeyIdentifier` seems good. What's your preference on `string` vs `bytes`?

haydentherapper: I'd prefer `bytes` too. We could reuse `LogId`, but we'd need to move the comment on `key_id` up to the `TransparencyLogInstance` message. The motivation for a separate type would be to keep the spec comments alongside the message, but I think it can still be clean if moved out to a higher level. WDYT?

woodruffw: > We could reuse LogId, but we'd need to move the comment on key_id up to the TransparencyLogInstance message. The motivation for a separate type would be to keep the spec comments alongside the message, but I think it can still be clean if moved out to a higher level. WDYT?

That seems reasonable to me!

haydentherapper: I actually like this structure, because we can now use `LogId` to house both `key_id` and a future `generic_log_id` or something like that if we move away from logs represented by their keys.

woodruffw (approved):
LGTM!

woodruffw (approved):

kommendorkapten (approved):

PR #287, opened 2024-04-05
python: add a py.typed marker

woodruffw:
Per PEP 561: this is the standard way to indicate that a Python package contains its own type hints, which helps downstream consumers (like sigstore-python) typecheck its own usage of these APIs.

(It'd be nice to get this out ASAP, but not required -- we can do it whenever the next patch release comes up to keep the versions in sync here 🙂)


gen/pb-python/sigstore_protobuf_specs/py.typed
woodruffw, line 1: NB: `flit` (our Python build backend) should automatically detect this file, so just adding it should be sufficient.

haydentherapper (approved):

PR #319, opened 2024-05-10
gen: bump patch versions

woodruffw:
A couple of useful changes have landed since the last patch release, so this bumps the patch versions where relevant 🙂


woodruffw:
(Needs changelog entries, adding those in a second.)

haydentherapper (approved):

Issue #336, opened 2024-05-28
How to represent `REQUIRED` on `oneof` fields.


kommendorkapten:
The best option for marking a `oneof` field as `required`; i.e that _exactly_ one of the fields must be set seems be in flux. Currently we are marking _all_ fields in the `oneof` as `required`, which  appeared to be the preferred way, [link](https://github.com/aip-dev/google.aip.dev/issues/299#issuecomment-1602958404). Marking _all_ fields as `required` seems to cause problems for Go reflection based field validation: https://github.com/sigstore/sigstore-go/issues/63#issuecomment-2132279557

According to [this issue](https://github.com/aip-dev/google.aip.dev/issues/1147), there is an alternative approach. Let's explore that and see if it solves the issues we are facing, and also the annotations per that issue is more visually appealing. There is also a [PR to disallow field_behavior on oneof](https://github.com/aip-dev/google.aip.dev/pull/1181) so updating makes sense.

PR #337, opened 2024-05-28
Use an alternative form to mark oneof fields as required

kommendorkapten:
#### Summary
Use an alternative and updated method to mark `oneof` fields as `REQUIRED`. 

Closes: https://github.com/sigstore/protobuf-specs/issues/336

#### Release Note
* Updated how `oneof` fields are marked as `REQUIRED` in the protobuf

#### Documentation
N/A


haydentherapper (dismissed):
Great find for the fix. Do we need to test this with all language clients?

haydentherapper:
Dismissing my review, seems like this isn't WAI according to https://github.com/sigstore/sigstore-go/issues/63#issuecomment-2135585754

kommendorkapten:
@haydentherapper yes, I did see in the other PR that this was fully WAI, I haven't had time to dig into this. My fear is that relying on annotations for verifying field behaviour may differ between clients :(

PR #380, opened 2024-08-09
sigstore_rekor: clarify inclusion_promise requirement

woodruffw:
This clarifies the (expected) requirements around `inclusion_promise` slightly. In particular, it clarifies that `inclusion_promise` is optional if and only if another source of signed time is present. If no other source of signed time is present, then an `inclusion_promise` is required and **MUST** be verified.

For cross-referencing, this is the part of the Client spec that suggests this behavior:

> Timestamping. Currently, the Transparency Service includes a timestamp in its response to the Signer. This timestamp comes from the Transparency Service’s internal clock, which is not externally verifiable or immutable. For this reason, a Signer SHOULD get their signatures timestamped. However, a Signer MAY choose to omit the timestamping step; in this case, the Signer MUST use the Transparency Service to provide a timestamp for the signature.

(NB: Like the other requirements on bundle formats/required fields, this requirement is for short-lived certificate instances of Sigstore, like the Public Good Instance. CC @haydentherapper for thoughts on if/how this can be better communicated -- I'm happy to add additional language here or in the `sigstore_bundle.proto` file!)


protos/sigstore_rekor.proto
```diff
@@ -104,8 +104,10 @@ message TransparencyLogEntry {
         int64 integrated_time = 4 [(google.api.field_behavior) = REQUIRED];
         // The inclusion promise/signed entry timestamp from the log.
         // Required for v0.1 bundles, and MUST be verified.
-        // Optional for >= v0.2 bundles, and SHOULD be verified when present.
-        // Also may be used as a signed timestamp.
+        // Optional for >= v0.2 bundles if another source of signed time
+        // is present.
```
haydentherapper, line 108: One could interpret "another source of ... time" to be current time, but "current time" wouldn't be signed. We could be very prescriptive and say "When verifying long-lived certificates, the client MAY choose to not require a signed timestamp and instead use the system clock." Thoughts?

woodruffw: That sounds good to me! Updating

woodruffw: Updated in https://github.com/sigstore/protobuf-specs/pull/380/commits/d9edb63b1146eb1ef96dcebb3fd38ee1cef585aa -- it now talks about a "suitable" source of time, which can be either another signed time source *or* the current system time when the certs are long-lived. LMKWYT!

haydentherapper (approved):
Perfect, thanks!

PR #383, opened 2024-08-22
added version to the signing config message.

kommendorkapten:
#### Summary
The `SigningConfig` message lacks a media type. This PR adds it. Note that I did not chose to re-order the fields, so the `media_type` is at position `5`. This does not matter in practice, more aesthetic. As there exist a release with `SigningConfig` we can't re-order the fields. I don't think anyone _is_ using it though, but this is protobuf best practices.


haydentherapper (approved):

haydentherapper:
Thanks!

PR #384, opened 2024-08-23
Fix typos in documentation of Bundle type

nmiyake:
#### Summary
Fixes spelling errors in documentation for `Bundle` type.

#### Release Note
NONE


woodruffw:
@nmiyake Thanks for these improvements! Could you run `make all` locally to propagate these doc fixes to the generated libraries as well?

nmiyake:
Ah yes, good catch.

Unfortunately, I'm using an Apple silicon machine and it looks like the `make` script is not fully compatible with `arm64` machines: `typescript`, `ruby`, and `rust` all fail with errors of the form `InvalidBaseImagePlatform: Base image namely/protoc-all@sha256:33e47b2aece23a282a9f8d03a193c063cdd4b1f60d427b148b3c449b51a3ba3c was pulled with platform "linux/amd64", expected "linux/arm64" for current build (line 1)`.

I might be able to look into this further later, but blocked at the moment due to this issue.

loosebazooka:
Oh maybe we should be using a multiplatform image for protoc

kommendorkapten:
I have an old PR to go over this, let's see if I can get that up to speed again. I don't remember the details on how running `amd64` on Apple silicon, but it works for me and I didn't perform any configuration. Anyway, we should attempt to run the correct image for the target architecture. 

I'll update this PR with the result of running `make all`.

woodruffw (dismissed):

woodruffw (approved):

PR #407, opened 2024-09-30
[DNM] python: generate pydantic models with betterproto

woodruffw:
Do not merge; this is a significant breaking change. Just pushing it up to play around with it.


woodruffw:
Yep! I had completely forgotten about this one.

PR #418, opened 2024-10-18
update go to 1.21 in go.mod

dmitris:
PR updates the `go` directive to `go 1.21` from `go 1.18` in `go.mod` file.  Both direct dependencies contain that version in their respective go.mod files:
* https://github.com/googleapis/go-genproto/blob/796eee8c2d537a6e5c8de22908f4facea945ebde/googleapis/api/go.mod#L3
* https://github.com/protocolbuffers/protobuf-go/blob/v1.35.1/go.mod#L3

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/sigstore/protobuf-specs
 
-go 1.22
+go 1.22.0
 
 require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20230726155614-23370e0ffb3e
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/sigstore/protobuf-specs
 
-go 1.18
+go 1.21
 
 require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20230726155614-23370e0ffb3e
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/sigstore/protobuf-specs
 
-go 1.21
+go 1.22
 
 require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20230726155614-23370e0ffb3e


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/protobuf-specs
 
-go 1.18
+go 1.21
```
cpanato, line 3: why not 1.22?

dmitris: I'd be fine even with 1.23 😄 - changed to 1.22 now

kommendorkapten:
What problem are you trying to solve with this? The go version directive is a bit confusing, but IIRC it refers to the minimum version need to to compile the module. 

if we are bumping the version to > 1.22, we must set the patch version too, as it won't work otherwise (i.e. 1.22.0).

dmitris:
> What problem are you trying to solve with this? The go version directive is a bit confusing, but IIRC it refers to the minimum version need to to compile the module.

Currently it is impossible to execute any go commands in the repository such as `go list ./...`:
```bash
$ go mod tidy && git diff
diff --git a/go.mod b/go.mod
index 0ac4c88..b66c03f 100644
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/sigstore/protobuf-specs

-go 1.18
+go 1.21
+
+toolchain go1.23.2

 require (
        google.golang.org/genproto/googleapis/api v0.0.0-20230726155614-23370e0ffb3e
```

I'm surprised the CI doesn't have any complaints - maybe we should add some check that would reveal problems like this one?

> if we are bumping the version to > 1.22, we must set the patch version too, as it won't work otherwise (i.e. 1.22.0).

👍  I changed the directive to `1.22.0`.

kommendorkapten (approved):

PR #461, opened 2025-01-08
Add warning when verifying with artifact digest

### LINKED ISSUE #444: Discussion on providing digest/alg for artifact verification
Originally the verification options supported two options:
1. Verify a DSSE envelope
2. Verify a message message signature + an artifact

This was changed in https://github.com/sigstore/protobuf-specs/pull/406 to be able to support some cases in the conformance tests when verifying against a message digest.

Having this as part of the standard verification path have me wondering if this is really a good idea. The first is that this does not seem to be compatible with Sigstore-js and [node.js verify implementation](https://nodejs.org/api/crypto.html#class-verify) which requires the complete message to verify. Pulling in a dependency for crypto verification is something we would like to avoid I think.

Also, this is not compliant with certain algorithms such as Ed25519 (this will fail today with Sigstore-go if an artifact digest is provided), LMS, ML-DSA and SLH-DSA to name a few. We haven't started to talk about PQC yet, but at some point we need too. All of the listed algorithms have pre-hash version, but I can't guess how popular/common those will be.

My third concern is about API security. Crypto APIs should make it hard to do the wrong thing, and having the Sigstore verification APIs accept a message digest is not optimal. Providing an artefact is more secure as we can guarantee in the Sigstore client that we are verifying against the expected digest. Low level APIs such as Go's `ecdsa.VerifyASN1` is an example where providing the digest makes sense, but for higher level APIs it can be a risk. 

When providing a bundle with a message signature, we explicitly forbid against using the (untrusted) digest for signature verification: https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L119

But as we are allowing digests for verification, I can see how this can be misused. Although not the same vulnerability but similar is [CVE-2022-36056](https://github.com/sigstore/cosign/security/advisories/GHSA-8gw7-4j42-w388) in Cosign, where the content of a Rekor entry wasn't verified to actually match the artifact in question. 

## Proposal

I can understand that it's desirable from the conformance test suite to have this feature, but for regular verification I would like to see an [comment](https://github.com/sigstore/protobuf-specs/blob/169f62fe52775c177460cb5a185304f791523bf6/protos/sigstore_verification.proto#L139) about that this is not recommended and an optional feature (it breaks Ed25519 today).
Possibly even rename the policy option https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/signed_entity.go#L444 to `WithArtifactDigestUnsafe` similar to when no signer identity is provided. 

If we also make this an optional feature, how could this be expressed in the conformance test suite?

cc @woodruffw @haydentherapper @bdehamer

haydentherapper:
This discourages using an artifact digest to verify an in-toto attestation, and mentions that it breaks Ed25519 verification.

Fixes #444


cpanato (approved):

kommendorkapten (approved):
Thanks!

PR #467, opened 2025-01-10
use Pydantic dataclasses for Python bindings

woodruffw:
~~WIP; this is experimental.~~

This should improve the quality/DX of the Python protobuf-specs a bit, by allowing them to be used as Pydantic dataclasses.

As a downside, this is an API breaking change for the Python bindings. However, sigstore-python strictly pins to each version so this shouldn't be a huge issue.


woodruffw:
> Will this affect sigstore-conformance?

It shouldn't directly -- `sigstore-conformance` pins to a version of `sigstore-python`, which itself fully pins `sigstore-protobuf-specs`. So this will only roll downwards after I make needed fixes on `sigstore-python` first 🙂

PR #481, opened 2025-01-15
bump ruby deps, protoc

bobcallaway:
I'm not 100% sure this is correct on the Ruby side of things, and I'm not quite sure how to best test this for breakage.


gen/pb-ruby/lib/sigstore_protobuf_specs/version.rb
```diff
@@ -16,6 +16,6 @@
 
 module Dev
   module Sigstore
-    VERSION = '0.3.3'
```
haydentherapper, line 19: Should we do 0.3.4? We have typically reserved 0.x bumps for breaking changes. 

We should also start looking into cutting a 1.x release after this, but that’s out of scope for now.

haydentherapper (approved):
discussed versioning offline

PR #492, opened 2025-01-16
add automation for keeping protoc and googleapis up to date

bobcallaway:
This creates automation to keep protoc and googleapis up to date. 

It configures dependabot to watch homebrew's build of protoc (via ghcr.io containers) to be notified of a new version being published (since that repo doesn't publish releases into any package manager).

The dependabot PR to bump the container reference will have a label added, which triggers a separate workflow to amend the PR with the correct changes to `protoc-builder/versions.mk`.

In addition, there's a monthly cron workflow that will fetch the latest commit hash from the https://github.com/googleapis/googleapis repo and update `protoc-builder/versions.mk` with it. This upstream repo doesn't use tags or cut releases at all.

### DIFF
--- a/.github/workflows/googleapis-update.yml
+++ b/.github/workflows/googleapis-update.yml
@@ -1,0 +1,40 @@
+name: Update Google APIs Commit Hash
+
+on:
+  schedule:
+    - cron: '0 0 1 * *'
+  workflow_dispatch:
+
+jobs:
+  update_protobuf_version:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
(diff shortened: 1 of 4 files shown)


cpanato (dismissed):
looks good

some non blocking

haydentherapper (approved):

.github/workflows/googleapis-update.yml
```diff
@@ -0,0 +1,40 @@
+name: Update Google APIs Commit Hash
+
+on:
+  schedule:
+    - cron: '0 0 1 * *'
```
loosebazooka, line 5: only you get pinged when these fail. Probably need to update this again with a group/bot account after submit.

bobcallaway: once https://github.com/sigstore/community/pull/535 merges I'll add a clause here to open an issue and tag folks

haydentherapper: This has merged

bobcallaway: i'll have to add the FYI feature back first.

PR #494, opened 2025-01-17
address zizmor issues, remove 3p action for creating release for java

bobcallaway:
This fixes issues reported by zizmor 🌈, as well as removing a dependency on a third-party action when the `gh` CLI should work just fine.

This also moves the gradle wrapper validation check to a non-deprecated action.

@loosebazooka @woodruffw fyi

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -7,4 +7,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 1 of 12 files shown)


loosebazooka (approved):
Java stuff looks good to me, ideally next step we add access to a pgp key to do the release to sonatype directly. We already have a java releasers group with permissions to write to the dev.sigstore on maven central

loosebazooka:
We already have a java releasers group with permissions to write to the dev.sigstore on maven central

bobcallaway:
> Java stuff looks good to me, ideally next step we add access to a pgp key to do the release to sonatype directly. We already have a java releasers group with permissions to write to the dev.sigstore on maven central

yup, i'll do that in a separate PR

woodruffw (approved):
LGTM on the `zizmor` side!

PR #501, opened 2025-01-19
fix googleapi update PR title/body, and remove comments

bobcallaway:
- Makes the PR title consistent with dependabot style and also refers to correct variable name.
- Runs `make all` after bumping the GOOGLEAPIS_COMMIT so that all relevant changes are included in the PR for review

### DIFF
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -15,44 +15,40 @@
 
 name: Check generated code for changes
 
-on: [push, pull_request]
+permissions:
+  contents: read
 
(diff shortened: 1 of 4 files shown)


cpanato (approved):

bobcallaway:
@cpanato now im confused... why does the filters on python-build.yml fire (due to the explicit match on `protoc-builder/versions.mk`, but the original filter of `protoc-builder/**` wouldn't cause the code gen steps to fire?

haydentherapper (approved):

PR #506, opened 2025-01-19
bump up to v0.4.0, add changelog

bobcallaway:
consistently sets all code at `v0.4.0` and once merged this commit can be tagged which will trigger the release process


loosebazooka:
I thought we were keeping the spec version and library version in sync? Although I dunno how that's gonna be reasonable long term

bobcallaway:
@haydentherapper thoughta?

kommendorkapten:
No, the messages and releases are versioned independently, see [RELEASE.md](https://github.com/sigstore/protobuf-specs/blob/main/RELEASE.md?plain=1#L48-L60)

> Two of the messages, [Bundle](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto) and [TrustedRoot](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto), are expected to be persisted and serialized to disk, and exchanged via other mechanisms such as the [Sigstore TUF root](https://github.com/sigstore/root-signing). Therefore they contain a media_type. The media types are versioned, and so they must be updated appropriately according to semver. Each message SHOULD be versioned independently and so MAY differ from the targeted release. The media type represents the version of the message, not the release. Note that the media type does NOT capture the patch version, only major/minor.

kommendorkapten:
Should we also remove the reference to JSON Schema in the release steps: https://github.com/sigstore/protobuf-specs/blob/main/RELEASE.md#json-schema

kommendorkapten (approved):
Thank you!

haydentherapper (approved):
Thanks!

woodruffw (approved):
LGTM, thanks @bobcallaway!

PR #510, opened 2025-01-23
automate publishing Java releases from GHA workflow

bobcallaway:
@loosebazooka can you add commits to this PR to fully switch over to publishing through the workflow instead of a manual run of the script?

### DIFF
--- a/.github/workflows/java-release.yml
+++ b/.github/workflows/java-release.yml
@@ -40,6 +40,24 @@ jobs:
           java-version: 11 
           distribution: 'temurin'
 
+      - name: Install Cosign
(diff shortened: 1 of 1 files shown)


loosebazooka:
Yup

loosebazooka:
So this *should* work, I need to run workflow dispatch against the secrets grabbing to see if it really does

bobcallaway:
wildcard matching isn't quite working yet, so DNM

haydentherapper (approved):
can't speak to the gradle changes though

loosebazooka:
Workload federation fixed? What's it look like?

PR #520, opened 2025-01-30
Add Result message to communicate the result of a verification

codysoyland:
#### Summary

This PR is an attempt at formalizing the output of a verifier using a new message: `dev.sigstore.verification.v1.Result`.

This initial form was extracted from sigstore-go, from the type [`VerificationResult`](https://github.com/sigstore/sigstore-go/blob/2e07b0bc1626495861d2ee55fa529ccf45250ed8/pkg/verify/signed_entity.go#L201-L207), with the following changes:

- I removed `VerifiedIdentity` as it seems to echo back the verification inputs and the matching values can all be found in `CertificateSummary`.
- Added `status` and `error` fields which may be used to communicate a verification failure.
- Added `UnverifiedData` which may be used to communicate if there were unverified TSA/Tlog signatures, as described in https://github.com/sigstore/sigstore-go/issues/48

Things to consider:

- Is this a good general shape to use?
- Should we add support for error codes?
- Any other data that should be included?
  - perhaps the full verified certificate chain?

I haven't generated docs/code yet to keep the review simple.

Fixes https://github.com/sigstore/protobuf-specs/issues/66

### DIFF
--- a/protos/sigstore_verification.proto
+++ b/protos/sigstore_verification.proto
@@ -162,3 +162,164 @@ message Input {
         // provided.
         optional Artifact artifact = 4;
 }
+
+// CertificateSummary contains verified metadata from the certificate
+message CertificateSummary {
+        // The issuer of the certificate
+        string certificate_issuer = 1;
+
+        // The Subject Alternative Name, or SAN, from the certificate
+        string subject_alternative_name = 2;
+
+        // The OIDC issuer. Should match `iss` claim of ID token or, in the case of
+        // a federated login like Dex it should match the issuer URL of the
+        // upstream issuer. The issuer is not set the extensions are invalid and
+        // will fail to render.
+        // OID 1.3.6.1.4.1.57264.1.8 and 1.3.6.1.4.1.57264.1.1 (Deprecated)
+        string issuer = 3;
+
+        // Deprecated
+        // Triggering event of the Github Workflow. Matches the `event_name` claim of ID
+        // tokens from Github Actions
+        // OID 1.3.6.1.4.1.57264.1.2
+        string githubWorkflowTrigger = 4;
+
+        // Deprecated
+        // SHA of git commit being built in Github Actions. Matches the `sha` claim of ID
+        // tokens from Github Actions
+        // OID 1.3.6.1.4.1.57264.1.3
+        string githubWorkflowSHA = 5;
+
+        // Deprecated
+        // Name of Github Actions Workflow. Matches the `workflow` claim of the ID
+        // tokens from Github Actions
+        // OID 1.3.6.1.4.1.57264.1.4
+        string githubWorkflowName = 6;
+
+        // Deprecated
+        // Repository of the Github Actions Workflow. Matches the `repository` claim of the ID
+        // tokens from Github Actions
+        // OID 1.3.6.1.4.1.57264.1.5
+        string githubWorkflowRepository = 7;
+
+        // Deprecated
+        // Git Ref of the Github Actions Workflow. Matches the `ref` claim of the ID tokens
+        // from Github Actions
+        // OID 1.3.6.1.4.1.57264.1.6
+        string githubWorkflowRef = 8;
+
+        // Reference to specific build instructions that are responsible for signing.
+        // OID 1.3.6.1.4.1.57264.1.9
+        string buildSignerURI = 9;
+
+        // Immutable reference to the specific version of the build instructions that is responsible for signing.
+        // OID 1.3.6.1.4.1.57264.1.10
+        string buildSignerDigest = 10;
+
+        // Specifies whether the build took place in platform-hosted cloud infrastructure or customer/self-hosted infrastructure.
+        // OID 1.3.6.1.4.1.57264.1.11
+        string runnerEnvironment = 11;
+
+        // Source repository URL that the build was based on.
+        // OID 1.3.6.1.4.1.57264.1.12
+        string sourceRepositoryURI = 12;
+
+        // Immutable reference to a specific version of the source code that the build was based upon.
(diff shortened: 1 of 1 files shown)


haydentherapper (commented):
This looks really good!

I think we should duplicate what you've created for `(Unverified)TimestampResult` for Fulcio and Rekor as well.

I don't think we need to include the cert chain for Fulcio or public key for Rekor, rather we should include some unique ID to correlate back to the trust root. For Fulcio, that could be the URI. For Rekor, that could be the log ID (URI will be unique in the future with Rekor v2).

For Rekor, we should specify what log proofs were unverifiable.

For Fulcio, we could also call this "CertificateAuthorityResult" to handle private PKI verification. We can include the list of cert subject key IDs.

protos/sigstore_verification.proto
```diff
@@ -162,3 +162,164 @@ message Input {
...
+message UnverifiedData {
+        repeated UnverifiedTimestampResult timestamps = 1;
+}
+
+// Result captures the result of the verification
+message Result {
+        // MUST be application/vnd.dev.sigstore.verificationresult+json;version=0.1
+        string media_type = 1;
+
+        // MUST be either "success" or "failure"
```
segiddins, line 307: should be an enum?

protos/sigstore_verification.proto
```diff
@@ -162,3 +162,164 @@ message Input {
...
+        string status = 2;
+
+        // SHOULD contain an error message if status == "failure"
```
segiddins, line 310: also, MUST be empty if status is success?

protos/sigstore_verification.proto
```diff
@@ -162,3 +162,164 @@ message Input {
...
+        string error = 3;
+
+        // If the bundle contains a DSSE, this result shonuld contain the
+        // deserialized and verified statement
+        string statement = 3;
```
segiddins, line 315: bytes instead of string?

protos/sigstore_verification.proto
```diff
@@ -162,3 +162,164 @@ message Input {
...
+        string subject_alternative_name = 2;
+
+        // The OIDC issuer. Should match `iss` claim of ID token or, in the case of
+        // a federated login like Dex it should match the issuer URL of the
+        // upstream issuer. The issuer is not set the extensions are invalid and
+        // will fail to render.
+        // OID 1.3.6.1.4.1.57264.1.8 and 1.3.6.1.4.1.57264.1.1 (Deprecated)
+        string issuer = 3;
+
+        // Deprecated
```
kommendorkapten, line 181: The deprecated OIDs, can we ignore them here? They were marked as deprecated two years go (almost to the date) https://github.com/sigstore/fulcio/pull/945.

If a specific client want's to support them, that's fine. But from a standard perspective, I think we should only focus on the current extensions.

protos/sigstore_verification.proto
```diff
@@ -162,3 +162,164 @@ message Input {
...
+
+        // Source repository visibility at the time of signing the certificate.
+        // OID 1.3.6.1.4.1.57264.1.22
+        string sourceRepositoryVisibilityAtSigning = 22;
+}
+
+message SignatureResult {
+        // If the bundle contains a public key, the public_key_id field will
+        // contain the public key ID
+        string public_key_id = 1;
```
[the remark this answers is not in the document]
kommendorkapten: minor but this is called hint in the `PublicKeyIdentifier` message: 
https://github.com/sigstore/protobuf-specs/blob/e943ce11bfbc53bd39035d2bc092a4de3c882c66/protos/sigstore_common.proto#L169

protos/sigstore_verification.proto
```diff
@@ -162,3 +162,164 @@ message Input {
         // provided.
         optional Artifact artifact = 4;
 }
+
+// CertificateSummary contains verified metadata from the certificate
+message CertificateSummary {
```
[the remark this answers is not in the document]
kommendorkapten: They serve different different purposes, the `CertificateIdentity` is for verifying, as it allows regexps to be set for the SAN.

And it also allows for setting the extensions needed during verification. I agree that clients knowing with OIDs to use os not optimal, but clients can implement helper function here to configure the verification.

protos/sigstore_verification.proto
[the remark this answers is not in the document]
kommendorkapten: Should these be a oneof:
```protobuf
message SignatureResult {
        oneof {
                string public_key_hint = 1;
                CertificateSummary certificate = 2;
        }
}

File: protos/sigstore_verification.proto (hunk 3 above)
Comment (kommendorkapten) [reply protos/sigstore_verification.proto:315]: I think the comments should be rephrased a bit. 
```
// If the bundle contains DSSE wrapped in-toto statement , this result contains the
// deserialized in-toto statement
```

As an in-toto statement is a JSON encoded object, i.e. a string.

File: protos/sigstore_verification.proto [hunk 7]
```diff
@@ -162,3 +162,164 @@ message Input {
...
+        // The issuer of the certificate
+        string certificate_issuer = 1;
+
+        // The Subject Alternative Name, or SAN, from the certificate
+        string subject_alternative_name = 2;
```
woodruffw, line 172: This might not be an issue in practice, but flagging: X.509 technically has multiple SANs, and I don't think any of the Sigstore specifications/documentation explicitly say that Sigstore leaf certificates only have a single SAN. In practice this is true, but we _might_ want to flag some documentation improvements as a follow-up to this to emphasize that, as a matter of X.509 profile, Sigstore-compatible CAs only issue EE certs with singular SANs 🙂

kommendorkapten: We probably should add the type for the SAN too?

PR #539, opened 2025-02-14
Update SigningConfig to specify API version and validity periods

### LINKED ISSUE #474: SigningConfig proto to have start dates?
When rotating keys for rekor (while doing v2 sharding), currently we would require two signing events for root signing
1. add new key to TrustedRoot for new current rekor 
2. rotate signers to the new rekor in SingingConfig

Root signing is a somewhat expensive process but we don't want the ecosystem to end up in a situation where verifiers slightly behind on time can't verify new signatures.

I think this can be solved by making SigningConfig more flexible to key rotations. Potentially by adding startTimes or some sort of time for when signers should start using it. Signers could go down the list of providers and pick the first one that is valid.

fyi: @jku

haydentherapper:
In order to faciliate clients gracefully handling breaking API changes, the SigningConfig will now include API versions for each of the service URLs so that clients can determine what services they are compatible with. Additionally, we've included validity periods which will be used to faciliate Rekor log sharding, when we spin up new log shards and distribute new key material.

Fixes #474


haydentherapper:
A few things to note:

* I kept the old url fields because I'm not sure if that will break sigstore-python, which supports ClientTrustConfig.
* That also means we need a new media type version, which I added.
* I've included comments for the selection process. For CA and OIDC URLs, the client must select **one** URL that matches its criteria. For Rekor and TSA URLs, the client must select **all** URLs that match its criteria. If this is confusing, lemme know and we can think about grouping things differently.
* Naming is hard because we've already used the abbreviated names I would have used for the pluralized fields....so the names are a bit verbose.

kommendorkapten (dismissed):
Looks good!

protos/sigstore_trustroot.proto
```diff
@@ -162,24 +164,96 @@ message SigningConfig {
...
+
+        // URLs to OpenID Connect identity providers.
+        //
+        // These URLs **MUST** be the "base" URLs for the OIDC IdPs, which clients
+        // should perform well-known OpenID Connect discovery against.
+        //
+        // Clients must select only Service with the highest API version
+        // that the client is compatible with and that is within its
+        // validity period. Clients should select the first Service
+        // that meets this requirement.
```
cmurphy, line 213: If the sort order is the same as for TrustedRoot https://github.com/sigstore/protobuf-specs/blob/e943ce11bfbc53bd39035d2bc092a4de3c882c66/protos/sigstore_trustroot.proto#L108-L109 then this could result in a client choosing a service that is valid but about to expire, when a later service in the list would be valid for longer.

protos/sigstore_trustroot.proto
```diff
@@ -162,24 +164,96 @@ message SigningConfig {
...
+        repeated Service openid_connect_provider_urls = 7;
+
+        // URLs to Rekor-compatible transparency logs.
+        //
+        // These URL **MUST** be the "base" URLs for the transparency logs,
+        // which clients should construct appropriate API endpoints on top of.
+        //
+        // Clients must select ALL Services with the highest API version
```
cmurphy, line 221: This seems like a big change, as clients don't currently upload entries to multiple logs at once, and may not always want to. Could this be changed to `ANY`?

protos/sigstore_trustroot.proto
```diff
@@ -162,24 +164,96 @@ message SigningConfig {
...
+        // should construct an appropriate CSR endpoint on top of.
+        // For example, if a CA URL is `https://example.com/ca`, then
+        // the client **MAY** construct the CSR endpoint as
+        // `https://example.com/ca/api/v2/signingCert`.
+        //
+        // Clients must select only Service with the highest API version
+        // that the client is compatible with and that is within its
+        // validity period. Clients should select the first Service
+        // that meets this requirement.
+        repeated Service certificate_authority_urls = 6;
```
cmurphy, line 203: I wish the new field names were more meaningfully different from the old names but I don't have a great suggestion.

protos/sigstore_trustroot.proto
```diff
@@ -148,9 +148,11 @@ message TrustedRoot {
 // signing. In particular, it primarily contains service URLs that a Sigstore
 // signer may need to connect to for the online aspects of signing.
 message SigningConfig {
-        // MUST be application/vnd.dev.sigstore.signingconfig.v0.1+json
+        // MUST be application/vnd.dev.sigstore.signingconfig.v0.1+json or
+        // application/vnd.dev.sigstore.signingconfig.v0.2+json
```
jku, line 152: This seems a bit unclear: after this PR this message documents `v0.2` only but that's not said anywhere. Do we allow v0.1 to be used in future as well?

Bundle uses this language:
>         // MUST be application/vnd.dev.sigstore.bundle.v0.3+json when encoded as JSON.
>         // Clients must to be able to accept media type using the previously defined formats:
>         // [list follows]

Of course situation here is a bit different: we should document 
* what we expect the instance to serve 
  * I'd like to say "MUST serve v0.2" but obviously there's a migration period here
* what we expect clients to handle
  * MUST handle v0.2, SHOULD handle v0.1

protos/sigstore_trustroot.proto
```diff
@@ -148,9 +148,11 @@ message TrustedRoot {
...
         string media_type = 5;
 
+        // Deprecated: Use certificate_authority_urls
```
jku, line 155: Can you explain the way these deprecations are useful (compared to just removing the fields):
* if a service does not want to support the new fields, I woudl expect them to just keep serving signingconfig v0.1? 
* if a client supports signingvonfig v0.2 then they need to support the new fields so won't need these

Maybe I'm missing a case?

jku: is the idea that clients have an easier time supporting v0.2 if they can just update version number and keep doing what they are doing now?

jku:
I guess one option to make the transition to new SigningConfig as easy as possible is to not replace the old one but just offer a new one:
* a new TUF artifact `signing_config2.json` is added alongside the existing `signing_config.json`
* the clients can implement support on their schedule: instead of parsing `signing_config.json` they download/parse `signing_config2.json` 
* at some point in future `signing_config.json` artifact is removed from TUF repository

This would be a way to avoid a flag day for SigningConfig

jku:
> I guess one option to make the transition to new SigningConfig as easy as possible is to not replace the old one but just offer a new one:

I'll have a better look at this again this week as I feel there is something to this idea -- we may be mistakenly combining the two problem cases (backwards incompatible service migration and plain log instance rotation), the right solutions to each  might actually be different

jku:
Okay, I couldn't actually make experiments with sigstore-python because of an ongoing protobuf-specs incompatibility, but I have tried to fully understand this and here's my current thinking:

### On versioning SigningConfig itself

* technically this PR makes only "compatible" changes to SigningConfig so we could update signing_config.json in root-signing without major issues.  root-signing artifacts should be considered API and changes should avoid flag days when possible so this seems good...  however:
* the moment we switch on a rekor v2 log and add it to the config in root-signing, this compatibility is shown to be fake: some clients seem to be compatible with the new signingconfig but that's only because they rely on the deprecated fields that won't contain the new log

Let's consider this approach:
* this pr actually makes incompatible SigningConfig changes (in other words the deprecated fields are removed)
* we add a new artifact into root-signing: "signing_config_N.json" (where N is the version), the old config remains as "signing_config.json"
* at this point clients can be updated to handle the new artifact (if we have already added the new artifact to both root-signing-staging and root-signing, they can remove the old signingconfig handling at the same time)
* at some point in future we fully deprecate old signing_config by removing the old artifact from root-signing

I think that might be best of both worlds: avoids a flag day but still makes it clear the API actually changes (and allows us to see from CDN logs how much the old signing_config.json is still used). comments @haydentherapper ?

### On Service.major_api_version

It's really hard to say if this will actually be a mechanism that works in cases that are not this specific upgrade (rekor v1->v2) but I can't figure out anything smarter. LGTM

protos/sigstore_trustroot.proto
haydentherapper: With the removal of the old fields, I've renamed them to the abbreviated service names

protos/sigstore_trustroot.proto
haydentherapper: Good call out. I've added the following:

```
     // All listed Services SHOULD be sorted by the `valid_for` window in
     // descending order, with the newest instance first.
```

With the newest instance first, this should limit how far into the list clients need to search to find a valid instance. Typically a client will just select the first from the list. With a sharding, a client will select the second (old shard) until the first in the list (new shard) becomes valid. For API changes, a client may have to search further into the list, but major API changes will be infrequent.

protos/sigstore_trustroot.proto
haydentherapper: Updated to serve only v0.2, and say clients can optionally handle v0.1. In practice, I assume all will drop support.

protos/sigstore_trustroot.proto
haydentherapper: I was aiming to concurrently support both to avoid a breaking change, but you're right, this isn't necessary. 

Also, only sigstore-python and -go support signing configs and I'm fairly certain no one is using them for the Go client, so making this breaking change is likely to have no impact.

protos/sigstore_trustroot.proto
haydentherapper: Originally, I was going to say that since this is for signing, I would expect that a user provide _all_ the logs that they want to contact. Thinking more on this, Certificate Transparency has a similar concept with log lists, and doesn't require its clients to write to all logs, only a subset.

I'm proposing now that we include a configuration for how clients should select Rekor and TSA instances (we don't need a configuration for Fulcio and the OIDC provider because the bundle only supports one certificate). The config will specify a selection of `ANY`, `ALL`, or `EXACT`, with another field `count` when `EXACT` is specified (like CAs writing to 2 logs based on browser requirements).

We need to decide on what we as the PGI operator will set. I think `ANY` is fine since I expect we'll only ship one log for now.

Thanks for this idea!

protos/sigstore_trustroot.proto
```diff
@@ -148,38 +148,127 @@ message TrustedRoot {
...
+        dev.sigstore.common.v1.TimeRange valid_for = 3;
+}
+
+// ServiceSelector specifies how a client should select a set of
+// Services to connect to. A client SHOULD throw an error if
+// the value is SERVICE_SELECTOR_UNDEFINED.
+enum ServiceSelector {
+        SERVICE_SELECTOR_UNDEFINED = 0;
+        ALL = 1;
+        ANY = 2;
```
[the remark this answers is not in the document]
kommendorkapten: I think it would be enough to document that it's up to the implementation tho decide how to chose the instances?

kommendorkapten (dismissed):
Nice work!

protos/sigstore_trustroot.proto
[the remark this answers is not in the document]
haydentherapper: Added a comment stating it's up to implementation with a few suggestions.

kommendorkapten (approved):
Thank you!

PR #606, opened 2025-04-17
DO NOT MERGE: Add new TLog type for rekor v2.

loosebazooka:
Update (4/24) Looks like we just go for option 2?

There are two options (both included in this commit)
1. Change the original rekor.v1.TransparencyLogEntry
- change logindex to uint
- make inclusionTime/Promise optional

2. Create a new rekor.v2.TransparencyLogEntry
- just a whole new type that's fully compatible with rekor v2
- a new repeated field in the bundle for v2 entries


loosebazooka:
I think we need to decide how to handle the differences in the TLogEntry, as mentioned here: https://github.com/sigstore/rekor-tiles/issues/147

The option where we create a new type feels a bit cleaner... but I think we should discuss it?

protos/sigstore_rekor_v2.proto
```diff
@@ -0,0 +1,112 @@
...
+// InclusionProof is the proof returned from the transparency log. Can
+// be used for offline or online verification against the log.
+message InclusionProof {
+        // The index of the entry in the tree it was written to.
+        uint64 log_index = 1 [(google.api.field_behavior) = REQUIRED];
+        // The hash digest stored at the root of the merkle tree at the time
+        // the proof was generated.
+        bytes root_hash = 2 [(google.api.field_behavior) = REQUIRED];
+        // The size of the merkle tree at the time the proof was generated.
+        int64 tree_size = 3 [(google.api.field_behavior) = REQUIRED];
```
haydentherapper, line 60: Can we make this unit64 too?

haydentherapper (commented):
Thanks for posting these solutions. My vote is for a new v2 rekor package. This looks like a clean solution and should be straightforward for clients to handle.

protos/sigstore_rekor_v2.proto
loosebazooka: yeah it should be.

protos/sigstore_rekor_v2.proto
```diff
@@ -0,0 +1,106 @@
...
+// signatures, but the first SHOULD be the signature from the log. Checkpoint contents
+// are concatenated with newlines into a single string.
+// The checkpoint format is described in
+// https://github.com/transparency-dev/formats/blob/main/log/README.md
+// and https://github.com/C2SP/C2SP/blob/main/tlog-checkpoint.md.
+// An example implementation can be found in https://github.com/sigstore/rekor/blob/main/pkg/util/signed_note.go
+message Checkpoint {
+        string envelope = 1 [(google.api.field_behavior) = REQUIRED];
+}
+
```
loosebazooka, line 50: These messages `KindVersion` and `Checkpoint` are repeated, they can probably be moved to a new common packaged (or be directly re-used from the rekor.v1 package). I don't think package information is used for anything but code gen -- but I Could be wrong. Either way moving it would require anyone importing the new builds to rewrite a little code.

protos/sigstore_rekor_v2.proto
```diff
@@ -0,0 +1,106 @@
...
+        Checkpoint checkpoint = 5 [(google.api.field_behavior) = REQUIRED];
+}
+
+// TransparencyLogEntry should directly capture the response from Rekor's
+// V2 API when creating a new entry in the log.
+message TransparencyLogEntry {
+        // The index of the log entry
+        uint64 log_index = 1 [(google.api.field_behavior) = REQUIRED];
+        // The unique identifier of the log.
+        dev.sigstore.common.v1.LogId log_id = 2 [(google.api.field_behavior) = REQUIRED];
```
haydentherapper, line 78: I'd like to propose that we drop log_id. It's an outdated reference to a log and assumes that every log has a unique signing key, which may not be true.

Clients should use the checkpoint's key ID to determine which log from TrustedRoot should be used. The key ID is computed based on the log origin/URL and the public key, reducing collisions.

Using the TrustedRoot, clients can either compute the key IDs for all logs themselves using the function from https://github.com/C2SP/C2SP/blob/main/signed-note.md#signatures, or just get it from `checkpoint_key_id` and treat the key ID as an opaque string.

A client could also use the checkpoint origin and fetch the log public key from the TrustedRoot by comparing against `base_url`, but I'd rather not assume uniqueness and use the key ID, which is the c2sp-spec compliant way to verify checkpoints.

loosebazooka: currently the tuple for determining which rekor instance to "verify" against is (logid, timestamp).  

So we're just switching it to (checkpoint_key_id, timestamp) ?

haydentherapper: Correct, and I also think we should compare base_url and the checkpoint origin, though arguably that's not strictly necessary since the origin is used to compute the key id.

Here's the proposal - https://github.com/sigstore/protobuf-specs/compare/main...haydentherapper:protobuf-specs:cpkeyid?expand=1
Deprecate log_id in the trusted root, but keep setting it for now in addition to checkpoint_key_id. As long as two logs don't use the same key, this will be fine.

haydentherapper:
Thanks @loosebazooka!

For the checkpoint key ID, there's still some ongoing discussions, but we will continue to populate log ID as it's effectively a unique identifier for the public good instance since it won't reuse the same signing key across shards.

PR #607, opened 2025-04-17
Add stronger language around how clients should handle service selection

haydentherapper:
This recommends that clients not select services across multiple API versions. I'm not sure the use case when we'd want them to select multiple across API versions, but for now, saying this is a SHOULD NOT rather than MUST NOT.


haydentherapper:
I also double checked the current language is sufficient for stating clients should only select from the highest API version supported.

cpanato (approved):

PR #610, opened 2025-04-21
Add section on adding new algorithms

haydentherapper:
This gives advice on how to add a new algorithm and makes the algorithm registry spec the source of truth.

I pulled in a little language from https://github.com/sigstore/sigstore/blob/main/POLICY.md, which was approved by the TSC, but will be updating this policy to also point to the alg registry spec.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,27 @@ $ make all
 
 to generate the Go and Python files under `gen/`.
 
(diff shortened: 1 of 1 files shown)


haydentherapper:
cc @arthurus-rex

woodruffw (approved):
Nice!

PR #629, opened 2025-05-01
Revise checkpoint key ID comment, deprecate log ID

haydentherapper:
Our use of log ID assumes that each deployment has a unique key. Additionally, its calculation was based on RFC6962 for CT.

For Rekor v2, we will now follow the C2SP spec to compute key IDs. For ECDSA keys, the log_id value will match. Ed25519 and RSA won't however. Rather than continue to set both of these values, I'd like to deprecate log_id and replace it with checkpoint_key_id for future log deployments.

Also update the base_url comment to state that it should match the log checkpoint origin, which will be true for Rekor v2. Rekor v1 has its own computation for an origin that isn't a valid URL.


protos/sigstore_trustroot.proto
```diff
@@ -43,22 +45,34 @@ message TransparencyLogInstance {
...
+        //
         // This is provided for convenience. Clients can also calculate the
         // checkpoint key ID given the log's public key.
-        // SHOULD be set for logs generating Ed25519 signatures.
         // SHOULD be 4 bytes long, as a truncated hash.
+        //
+        // To find a matching TransparencyLogInstance in the TrustedRoot,
+        // clients will parse the checkpoint, and for each signature line,
+        // extract the key name (i.e. log origin, base_url) and 4-byte key ID,
+        // and compare the key name and key ID against each log instance.
```
jku, line 75: maybe avoid saying that key name is compared since it seems only the keyid is compared:
```suggestion
        // use the key name (i.e. log origin, base_url) and public key to calculate
        // key ID which can then be compared against the TrustedRoot log instances
```

jku (approved):
Makes sense to me, left one improvement suggestion but it works as is

protos/sigstore_trustroot.proto
haydentherapper: Thanks, incorporated!

haydentherapper: Actually sorry, I need to revert this partially. The client wouldn't know the log public key yet. The client only has what the checkpoint provides, which is the origin and key ID in the checkpoint signature line. The client can verify that the key ID is properly computed if it wants to after it retrieves the public key from the TrustedRoot, but I don't think that's strictly necessary.

kommendorkapten (approved):

PR #634, opened 2025-05-08
Add operator for SigningConfig services, log and TSA roots

haydentherapper:
When verifying a log entry or timestamp with a threshold, clients should verify that log proofs or signed timestamps are from distinct operators, not just distinct deployments. For example, if an operator deploys multiple log services, either due to multiple versions or just redundancy, multiple valid proofs in verification material should be treated as a single valid proof for the threshold.

Since verification will require distinct operators, we also need to include operators for the signing config, so that when multiple services are picked, the client will only select distinct services so that verification will succeed.


haydentherapper:
This adds a little more complexity to the selection logic for SigningConfig, but I think the tradeoff is worth it, since we _need_ to group log deployments by operators to meet thresholds.

kommendorkapten (approved):
Thanks!

kommendorkapten (changes requested):
We should bump the version of the trusted root media type to v0.2 as we are adding new fields. 

Also, should we describe any formats for the operator? The name of the entity/organization? URI? My recommendation would be a schema-less URI, avoiding any path components if possible. e.g. `sigstore.dev`, `github.com` etc.

protos/sigstore_trustroot.proto
```diff
@@ -87,6 +93,12 @@ message CertificateAuthority {
         // The TimeRange should be considered valid *inclusive* of the
         // endpoints.
         dev.sigstore.common.v1.TimeRange valid_for = 4;
+        // The name of the operator of this certificate authority. This MUST
+        // be used when there are multiple timestamp authorities to determine
+        // if the signed timestamp verification meets a specified threshold,
+        // e.g. two signed timestamps from timestamp authorities operated by
+        // the same operator should count as only one valid timestamp.
+        string operator = 5;
```
loosebazooka, line 101: I think this part is maybe a bit confusing, because we re-use CertificateAuthority for TSAs.

Should we switch TSA to it's own type?

protos/sigstore_trustroot.proto
```diff
@@ -231,7 +243,8 @@ message SigningConfig {
 // Clients MUST use the API version hint to determine the service with the
 // highest API version that the client is compatible with. Clients MUST also
 // only connect to services within the specified validity period and that has the
-// newest validity start date.
+// newest validity start date. Clients MUST select services from distinct operators
+// when selecting multiple services.
```
loosebazooka, line 247: I think we should provide a suggested algorithm for selection, it's starting to get pretty confusing.

haydentherapper:
@kommendorkapten We didn't bump the trusted root version when we introduced checkpoint key ID awhile ago, clients would just ignore that field. Wouldn't we be breaking the verification path for clients who support only the v0.1 trusted root format? We could instead simply document this as a required change to start reading this field as part of supporting the verification of Rekor v2/TLE v2.

Agreed on describing the format, I'll use a schema-less URI.

haydentherapper:
Though, with the recent update to the architecture docs, we have specified that we will version trusted root and signing config. We could ship both the v1 and v2 versions.

protos/sigstore_trustroot.proto
kommendorkapten: I'm not sure it would be more clear if the TSA was it's own type? It would be a pretty big change for the clients if we did. I'm inclined to keep as is.

haydentherapper: Agreed, if we were to revisit this and cut a new version, then I'd be more inclined, but for now, I'd keep it as is. I'll update the comment a bit to acknowledge this. Also, having an operator for a CA would allow a client to potentially filter out operators it doesn't want to use.

loosebazooka (approved):

haydentherapper:
Once merged, I'll take care of updating sigstore-go.

kommendorkapten (approved):
Thanks for this!

PR #643, opened 2025-05-20
Add service generation image

loosebazooka:
This should be published to ghcr to be re-used by rekorv2, fulcio and any other services we define in proto/grpc

#### Summary
An sigstore internal use image for generating proto/grpc files from proto definitions in our service projects. A follow-up PR will be required to make this available on ghcr for sharing so it can be consumed in other repos.

#### Release Note
none

#### Documentation
none


Makefile
```diff
@@ -45,6 +46,14 @@ go: base-image-go
 	  -I/opt/include -I/googleapis -I/defs/protos \
 	  --go_opt=module=github.com/sigstore/protobuf-specs/gen/pb-go --go_out=/defs/gen/pb-go ${PROTOS}
 
+# an image on ghcr for generating defintions for sigstore services
+services-image:
```
bobcallaway, line 50:
```suggestion
services-image: go
```
should we make the dependency more explicit?

haydentherapper (approved):
Thanks!

PR #662, opened 2025-06-05
workflows: Update python versions to currently supported

jku:
as per https://devguide.python.org/versions/


loosebazooka:
yeah whatever, we can do it wholesale in another PR across languages

PR #690, opened 2025-07-07
Introduce v0.2 TrustedRoot, un-deprecate log ID

haydentherapper:
As we learned when we introduced a new SigningConfig version, adding fields to the proto isn't a breaking change, but setting these fields in JSON and unmarshalling the object is a breaking change. Since we've added operator and checkpoint_key_id to the Instance message, this requires introduced a new TrustedRoot version, which I've done in this PR as v0.2. As with SigningConfig, I've made it optional for clients to indefinitely support old versions, but it should be little work for a client to support both versions concurrently.

I've also un-deprecated log ID. This was premature since no clients supported checkpoint key ID yet. With Rekor v2, we'll be setting the same value in both log_id and checkpoint_key_id, so clients can continue to ignore checkpoint_key_id for now.


gen/pb-python/sigstore_protobuf_specs/dev/sigstore/rekor/v2/__init__.py
```diff
@@ -67,50 +67,50 @@ class Signature(betterproto.Message):
 
 
 @dataclass(eq=False, repr=False)
-class DsseRequestV002(betterproto.Message):
```
haydentherapper, line 70: @woodruffw any guesses as to why my build output differs from CI? I can just omit this file, I’ve seen this before though.

kommendorkapten (dismissed):
Changes looks good, thanks.

kommendorkapten (approved):

haydentherapper:
Do y'all have opinions on if I should cut a new minor release vs a patch release? Given this is a new TrustedRoot release, I'm leaning towards a new minor 0.5

kommendorkapten:
A new minor seems like the correct choice 👍

PR #705, opened 2025-07-22
Add required annotations for Service message in SigningConfig

haydentherapper:


Makefile
```diff
@@ -32,10 +32,10 @@ DOCKER_BUILD = docker build --platform ${PLATFORM} --build-arg UID=${UID}
 DOCKER_RUN = docker run --platform ${PLATFORM} --user ${UID}:${GID}
 
 # base protos for clients that do not want to include service-protos
-BASE_PROTOS = $(shell find protos/ -iname "*.proto" | sed 's|^|/defs/|')
+BASE_PROTOS = $(shell find protos/ -iname "*.proto" | sed 's|^|/defs/|' | sort)
 BASE_PROTO_INCLUDES=-I/opt/include -I/googleapis -I/defs/protos
 
-PROTOS = $(shell find protos/ service-protos/ -iname "*.proto" | sed 's|^|/defs/|')
+PROTOS = $(shell find protos/ service-protos/ -iname "*.proto" | sed 's|^|/defs/|' | sort)
```
haydentherapper, line 38: @woodruffw @jku do you know why the order of the protos would matter when generating the python files? I finally found the root cause of the difference in generated code - `find` would return the protos in a different order on macOS vs Linux.

Issue #729, opened 2025-08-12
Order of instances in signing config.


kommendorkapten:
In trusted root, the instances are ordered in ascending order: 

https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto#L144-L150

In the signing config, the order is the opposite descending order:

https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto#L208-L209

Is there any reason for deviating from the ascending order for signing config? I would prefer if we are consistent with ascending order.

@haydentherapper @loosebazooka @jku

jku:
Alternatively, maybe there is no reason to say anything about trustedroot log order -- or am I missing a reason why it matters?

PR #732, opened 2025-08-13
Relaxed the requirement of only one instance allowed to have open timespan

kommendorkapten:
Closes https://github.com/sigstore/protobuf-specs/issues/729

#### Summary

This requirement causes problems with running multiple instances as can be specified in the signing config. For a client's perspective, this does not really matter as not having an end-time is similar to having an end time that's set in the _far future_. As clients must work with overlaps from instances, this should not be a breaking change.

#### Release Note

* Removed the requirement of only allowing a single instance in the trusted root to have an open timespan for the validity time.

#### Documentation

N/A


haydentherapper (approved):
Thank you!

PR #734, opened 2025-08-17
move to protoc v32.0, re-enable dpbot automation

bobcallaway:
this had previously been disabled because homebrew was having issues (we don't use the content from homebrew, but we use it as a trigger to drive updates).


loosebazooka:
Which part re-enables automation?

loosebazooka (approved):

PR #735, opened 2025-08-17
move to trixie from bullseye

bobcallaway:
shouldn't have any effect on codegen, just uses more up to date versions of utilities


loosebazooka:
Distroless hasn't yet moved to Trixie if this might cause a glibc compat issue

bobcallaway:
> Distroless hasn't yet moved to Trixie if this might cause a glibc compat issue

fair point, we can hold on this until distroless moves

loosebazooka (approved):

PR #860, opened 2026-03-25
Add ML-DSA-44 algorithm identifier

Hayden-IO:
From discussions with cryptographers, the general consensus is ML-DSA-44 is sufficient for PQC signing, with smaller keys and signatures. ML-DSA-65 and ML-DSA-87 will be primarily for specialized use cases, e.g. gov't requirements.
Additionally, the witness network is likely to use ML-DSA-44.

Updated the comment for LMS/LMS-OTS to state it should not be used at all, as there are no clients that will support this.


kommendorkapten:
This looks good, would you also update https://github.com/sigstore/architecture-docs/blob/main/algorithm-registry.md

Re: SHAKE-256, How do you feel about not adding that now. Reading FIPS 204 $5.4, hash algorithms specified by FIPS 180-4 with sufficient output size (e.g. SHA2-256) is allowed, so maybe punt on adding SHAKE-256 until we _need_ it? We already have a lot of algorithms listed (for ML-DSA lambda is 128 bits).

From FIPS 204 $5.4:

> In order to maintain the same level of security strength when the content is hashed at the application level
or using HashML-DSA , the digest that is signed needs to be generated using an approved hash function
or XOF (e.g., from FIPS 180 [8] or FIPS 202 [7])

But also, if we can work towards always hashing on the application layer (e.g. use in-toto) there is no need for the pre-hash version, and we can stick with pure only. E.g. newer cosign only signs an in-toto statement for OCI images. We could do the same for blobs and so only sign the in-toto statement referencing the blob.

Hayden-IO:
https://github.com/sigstore/architecture-docs/pull/61 for the registry update

Issue #249, opened 2024-03-06
Trust root: switch to a trust "pool" over a set of CA chains


woodruffw:
This would be a full generalization of #245: the `TrustRoot` as currently specified contains a 2D list (a list of CAs, each with one or more cert "chains"), which is now how X.509 talks about path validation (path validation in X.509 is an abstract graph construction, ending at *any* trust anchor regardless of whether it's self-signed or not).

As a concrete proposal: `repeated CertificateAuthority certificate_authorities = 3;` would be replaced by `repeated X509Certificate trust_anchors` or similar. This would remove some of their fields that are currently present in the `CertificateAuthority` message, but AFAICT these are mostly duplicates of X.509 state anyways (e.g. the subject and validity period).

From here, the semantics of `trust_anchors` would be "first chain wins," i.e. the trust distributor can decide whether to put an intermediate or root in the trust anchors, and clients are (per X.509) perfectly entitled to terminate chain construction as soon at they reach the trust anchor, even if the member is not a self-signed root. This would essentially imply the same thing as #245, but without the explicit ordering (since there is no true ordering in a cert pool).

CC @haydentherapper @jku @loosebazooka 🙂

woodruffw:
> Would we lose the other metadata in CAs? Validity period doesn't necessarily need to march up with what is provided by the certificate. We could choose to deprecate it earlier than the certificate defined validity period.

Ah, I hadn't thought of this. To make sure I understand: is the value of an external validity that the trust root might still want to distribute a compromised CA, but limit its applicability to the period _before_ the compromise? That makes sense if so, and seems like a good thing; I have some ideas for equivalent approaches if that's what it's needed for 🙂 

> Url is also something we use in sigstore Java to define our signers

Could you share an example of this? I don't think we look at the URL at all in sigstore-python, I'm wondering if we're missing something.

haydentherapper:
In general I'm supportive of this, though Appu's question about keeping track of validity needs to be solved. Two thoughts:

Can we also have `repeated X509Certificate untrusted_certs` or `repeated X509Certificate chain_builders` so that we can distribute intermediates alongside trust anchors?

Besides the validity question, is there a benefit in keeping the existing field and making this a one-of? Providing the exact chain could be an performance optimization to avoid chain building, for example where you might have only one valid path through a bunch of intermediates.

[...]

PR #898, opened 2026-06-23
WIP: Bundle Version V2

kommendorkapten:
#### Summary
This is a draft PR to discuss how the bundle v2 format could look like to prepare for the PQC migration.


protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,99 @@
...
+        bytes proof = 1 [(google.api.field_behavior) = REQUIRED];
+        // If the artifact was signed prior to log submission.
+        // Signature is over
+        // c2sp.org/identity-transparency/v1 || 0x00 || SHA256(SHA256(data))
+        optional dev.sigstore.common.v1.PublicKeyIdentifier public_key = 2;
+}
+
+message PublicKey {
+        dev.sigstore.common.v1.PublicKeyIdentifier public_key = 1 [(google.api.field_behavior) = REQUIRED];
+}
```
steiza, line 73: Were we going to have optional `TimestampVerificationData` here?

kommendorkapten: Yes, they were removed by mistake in https://github.com/sigstore/protobuf-specs/pull/898/changes/c663e917455102c766fe032d1a95dc2922c9c73a I'll add them back.

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,100 @@
...
+        // if message is submitted to a log
+        bytes signature = 1;
+}
+
+message Embedded {
+        string payload_type = 2;
+        bytes payload = 2;
+        // The PAE used before signing is:
+        // "DSSEv1" + SP + LEN(type) + SP + type + SP + LEN(body) + SP + body
+        // When signing, data is the PAE.
```
Hayden-IO, line 39:
```suggestion
        // When signing, data is the PAE.
        // Unset when submitting an identity token to the log.
```

Hayden-IO: To be explicit that this is empty in the OIDC case

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,100 @@
...
+        optional Signature = 3;
+}
+
+message MessageSignature {
+        dev.sigstore.common.v1.HashOutput message_digest = 1 [(google.api.field_behavior) = REQUIRED];
+        // When signing data is the artifact
```
Hayden-IO:
```suggestion
        // When signing, data is the artifact.
        // Unset when submitting an identity token to the log.
```

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,100 @@
...
+        optional Signature signature = 2;
+}
+
+message TlogProof {
+        // The C2SP proof.
```
Hayden-IO:
```suggestion
        // The log proof, formatted per c2sp.org/tlog-proof.
```

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,100 @@
...
+        // If OIDC based submit is used, the `extra` field contains the
+        // relevant claims needed for recreation of the leaf during
+        // verification. They are stored as an unordered list of elements
+        // matching key:value separated by `\n` (0x0a). Both key and value
+        // are strings.
+        // If a signature is part of the leaf, verifiers MUST verify the
+        // the artifacts signature, and compare that the signature matches
```
Hayden-IO:
```suggestion
        // the artifact's signature, and compare that the signature matches
```

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,100 @@
...
+        bytes proof = 1 [(google.api.field_behavior) = REQUIRED];
+        // If the artifact was signed prior to log submission.
+        // Signature is over
+        // c2sp.org/identity-transparency/v1 || 0x00 || SHA256(SHA256(data))
+        optional dev.sigstore.common.v1.PublicKeyIdentifier public_key = 2;
+}
+
+message PublicKey {
+        dev.sigstore.common.v1.PublicKeyIdentifier public_key = 1 [(google.api.field_behavior) = REQUIRED];
+        TimestampVerificationData timestamps = 2 [(google.api.field_behavior) = REQUIRED];
```
Hayden-IO:
```suggestion
        // Enables key rotation when signature is not logged
        optional TimestampVerificationData timestamps = 2 [(google.api.field_behavior) = REQUIRED];
```

Hayden-IO: Adding a comment for clarifying the purpose, and making it optional. We shouldn't always require a timestamp, otherwise it means you must depend on infrastructure to sign.

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,100 @@
...
+        TimestampVerificationData timestamps = 3 [(google.api.field_behavior) = REQUIRED];
+}
+
+message VerificationMaterial {
+        oneof content {
+                // Signed with a raw key pair
+                PublicKey public_key = 1 [(google.api.field_behavior) = REQUIRED];
+                // Signed with cert + rfc3161 timestamp
+                Cert cert = 2 [(google.api.field_behavior) = REQUIRED];
+                // Pushed to Rekor V3.
```
Hayden-IO:
```suggestion
                // Logged to Rekor V3, authenticating to the log with an
                // identity token or public key/signature
```

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,100 @@
...
+        dev.sigstore.common.v1.PublicKeyIdentifier public_key = 1 [(google.api.field_behavior) = REQUIRED];
+        TimestampVerificationData timestamps = 2 [(google.api.field_behavior) = REQUIRED];
+}
+
+message Cert {
+        oneof cert_data {
+                dev.sigstore.common.v1.X509Certificate certificate = 1 [(google.api.field_behavior) = REQUIRED];
+                dev.sigstore.common.v1.X509CertificateChain certificate_chain = 2 [(google.api.field_behavior) = REQUIRED];
+        }
+        TimestampVerificationData timestamps = 3 [(google.api.field_behavior) = REQUIRED];
```
Hayden-IO, line 82: Just to state it explicitly, I like that we make this required. We've supported certificates without timestamps, because you could deploy your own PKI that doesn't conform to the certificate profile and issues long-lived certificates, but this added complexity around verification.

Hayden-IO (commented):
Just a few comment updates to sync with the doc! Otherwise, LGTM!

kommendorkapten:
Thanks for feedback, updated now.

Hayden-IO:
@kommendorkapten Could you run make all? I'd like to be able to reference the generated code in a client change.

kommendorkapten:
The history looks terrible after I rebased onto main, will fix that.
I have ran `make all`, not sure if it will make any difference (at least not for go) but I had to change the architecture to `linux/arm64` as the docker build step was crashing.

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,105 @@
...
+message MessageSignature {
+        dev.sigstore.common.v1.HashOutput message_digest = 1 [(google.api.field_behavior) = REQUIRED];
+        // When signing, data is the artifact.
+        // MUST be unset when submitting with an ID token to a log.
+        optional Signature signature = 2;
+}
+
+message TlogProof {
+        // The log proof, formatted per c2sp.org/tlog-proof.
+        // If OIDC based submit is used, the `extra` field contains the
```
loosebazooka, line 53: should we "TBD" that this may be hashes or raw information?

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,105 @@
...
+
+message PublicKey {
+        dev.sigstore.common.v1.PublicKeyIdentifier public_key = 1 [(google.api.field_behavior) = REQUIRED];
+        // Optional timestamp to enable key rotation when a signature is not
+        // logged (witnesses provides a counter-signature with timestamp).
+        optional TimestampVerificationData timestamps = 2;
+}
+
+message Cert {
+        oneof cert_data {
```
loosebazooka: do we make any assertions about the cert here? Should we say must include trust anchor? Or must only be the untrusted portion? Do we care?

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,105 @@
...
+                // Signed with cert + rfc3161 timestamp
+                Cert cert = 2 [(google.api.field_behavior) = REQUIRED];
+                // Logged to Rekor V3, authenticating with an ID token or
+                // a public key/signature.
+                TlogProof tlog_proof = 3 [(google.api.field_behavior) = REQUIRED];
+        }
+}
+
+message Bundle {
+        // MUST be application/vnd.dev.sigstore.bundle.v2.0+json when
```
loosebazooka, line 98: should we intentionally mark this with pqc somewhere or is that weird?
like `vnd.dev.sigstore.bundle.pqc...`

protos/sigstore_bundle_v2.proto
Hayden-IO: Yes, this is underspecified in the current comment. In the doc I said it was digests. I'm strongly leaning towards the unhashed provenance values now. I know at least PyPI relies on them.

protos/sigstore_bundle_v2.proto
Hayden-IO: Yea, we should match the comment we have for the current bundle. We should discourage using a certificate chain, and if the chain is specified, it must not contain a trust anchor. And the certificate must match Fulcio's certificate profile

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,105 @@
...
+        optional Signature signature = 2;
+}
+
+message TlogProof {
+        // The log proof, formatted per c2sp.org/tlog-proof.
+        // If OIDC based submit is used, the `extra` field contains the
+        // relevant claims needed for recreation of the leaf during
+        // verification. They are stored as an unordered list of elements
+        // matching key:value separated by `\n` (0x0a). Both key and value
+        // are strings.
```
Hayden-IO, line 57:
```suggestion
        // are strings.
        // The key is the name of the context category as specified in the leaf specification.
        // When the credential is an identity token, all value strings will be the parsed identity
        // and provenance values from the token. When the credential is a public key,
        // the optional value will be the provided context as a hex-encoded digest.
```

Hayden-IO: @loosebazooka Here's what I'm going with thus far. Might be a change for the key-based context string, but that shouldn't impact any of the default workflows.

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,105 @@
...
+
+// Note that the DSSE Pre Auth Encoding (PAE) is used, but the signature
+// scheme is Sigstore specific. The raw message is never signed, it's
+// always hashed as there is a risk otherwise to transmit large payloads
+// to an HSM.
+message Signature {
+        // The signature is computed over a domain separated message:
+        // sigstore.dev/v1 || 0x00 || SHA256(SHA256(data)) if message is not
+        // submitted to a log
+        // c2sp.org/identity-transparency/v1 || 0x00 || SHA256(SHA256(data))
```
Leont, line 29: Should the digest be hard-coded to SHA-256? Maybe we should build in some flexibility just in case?

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,105 @@
...
+        // if message is submitted to a log
+        bytes signature = 1;
+}
+
+message Embedded {
+        string payload_type = 1;
+        bytes payload = 2;
+        // The PAE used before signing is:
+        // "DSSEv1" + SP + LEN(type) + SP + type + SP + LEN(body) + SP + body
```
Leont, line 38: This line should probably use the same names as the fields above? (`payload_type` and `payload`)

protos/sigstore_bundle_v2.proto
woodruffw: SHA2/256 is still a very conservative general purpose cryptographic digest in 2026, so IMO there isn't much benefit in malleability here.

(If we're worried about a distant future in which the SHA2 family is discovered to have weaknesses, then I think the appropriate thing for us to do is version the entire format rather than introduce parametricity within a version. In this case I think that means we'd do `sigstore.dev/v2` with the new fixed selection.)

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,105 @@
...
+// scheme is Sigstore specific. The raw message is never signed, it's
+// always hashed as there is a risk otherwise to transmit large payloads
+// to an HSM.
+message Signature {
+        // The signature is computed over a domain separated message:
+        // sigstore.dev/v1 || 0x00 || SHA256(SHA256(data)) if message is not
+        // submitted to a log
+        // c2sp.org/identity-transparency/v1 || 0x00 || SHA256(SHA256(data))
+        // if message is submitted to a log
+        bytes signature = 1;
```
Leont, line 31: IME a signature should be explicit about what signing scheme is being used. I think we should add a field for that.

protos/sigstore_bundle_v2.proto
Hayden-IO: In addition, the c2sp.org/tlog-tiles specification hardcodes SHA-256, for the same reason @woodruffw noted. If a weakness were found, we would have to create a new log revision anyways and update client libraries.

protos/sigstore_bundle_v2.proto
woodruffw: I might be misunderstanding, but I'm pretty sure the version of the overall format implies the signature algorithm. In this case I think it's always Ed25519, based on tlog-tiles.

(Like SHA2 above IMO this is considered best practice -- schemes should be versioned at the top level rather than encoding malleability within each version.)

Hayden-IO: Actually, it could vary, as both Ed25519 and ML-DSA (-44, though potentially all 3 NIST parameter sets, not sure if we want to support all yet) could be supported*.

Currently, signing scheme has been derived from the public key based on well-defined mappings between curves and digest algorithms. This isn't a concern for Ed25519 and ML-DSA which handle hashing as part of the scheme. Assuming we do want to support the 3 parameter sets for ML-DSA, either we need to derive the signing scheme based on public key size, or be explicit.

*Note that this bundle format was designed specifically for PQC. I think clients should just support ML-DSA-44, but I think the bundle format can be more flexible for private deployments.

protos/sigstore_bundle_v2.proto
Leont: That sounds reasonable (though I would point out that currently the trusted root is doing the opposite here by explicitly specifying the transparency roots use SHA-256).

Hayden-IO: Are you referring to the [log ID encoding](https://github.com/sigstore/protobuf-specs/blob/e21dd61cb740d8e1a1370060dac7a378271fcb59/protos/sigstore_trustroot.proto#L44-L46) or [checkpoint key ID](https://github.com/sigstore/protobuf-specs/blob/e21dd61cb740d8e1a1370060dac7a378271fcb59/protos/sigstore_trustroot.proto#L57-L81) (the latter is the recommended way to reference a log, since it's a unique identifier given the log's name/origin and key)?

These use SHA-256 as well to compute the IDs.

Leont: I mean the [hashAlgorithm](https://github.com/sigstore/root-signing/blob/main/targets/trusted_root.json#L6) field in transparency log entries.

Hayden-IO: Ah, yes, we should remove that field. We haven't had a strong motivation for a new revision of the trust root format, though there's a few fields I'd like to remove (like `logId`).

Originally, the thought was we could support different hash algorithms and that field gave us that flexibility, but now having built Rekor on top of transparency specifications, it's unnecessary.

protos/sigstore_bundle_v2.proto
Leont: Even if it may be clear now, it might not be in the future. A good example is RSA, where originally PKCS1 v1.5 padding was the norm but later PSS was introduced as a cryptographically stronger option.

Being explicit about the scheme, but only allowing a very limited set of values is the best way to handle that IME.

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,105 @@
...
+        // matching key:value separated by `\n` (0x0a). Both key and value
+        // are strings.
+        // If a signature is part of the leaf, verifiers MUST verify the
+        // the artifact's signature, and compare that the signature matches
+        // whats submitted to the log.
+        bytes proof = 1 [(google.api.field_behavior) = REQUIRED];
+        // If the artifact was signed prior to log submission.
+        // Signature is over
+        // c2sp.org/identity-transparency/v1 || 0x00 || SHA256(SHA256(data))
+        optional dev.sigstore.common.v1.PublicKeyIdentifier public_key = 2;
```
Hayden-IO, line 65: From prototyping with sigstore-go: I think we shouldn't include this identifier, in keeping with simplicity.

* I'm not sure if any clients used the hints. Our client interfaces currently take only a single key to validate a signature as well.
* The 'hint' is not well-defined. We suggest it should be SHA256(SubjectPublicKeyInfo) from RFC6962. That RFC is now irrelevant to this design.
* From chatting with rgdd about the leaf spec, an ask was to not use ASN.1 encodings for public keys.

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,105 @@
...
+        bytes payload = 2;
+        // The PAE used before signing is:
+        // "DSSEv1" + SP + LEN(type) + SP + type + SP + LEN(body) + SP + body
+        // When signing, data is the PAE.
+        // MUST be unset when submitting with an ID token to a log.
+        optional Signature signature = 3;
+}
+
+message MessageSignature {
+        dev.sigstore.common.v1.HashOutput message_digest = 1 [(google.api.field_behavior) = REQUIRED];
```
Hayden-IO, line 45: From prototyping with sigstore-go: Is this needed?

Similarly with the public key hint, it's useful if the interface supported verifying multiple artifacts as once, but that isn't how the APIs have been designed. If it's ignored, I'd suggest dropping it.

Also, is a little unclear which digest it is. SHA256(data)? Internal SHAKE256(message)? Unless we're clearly defining it, drop?

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,105 @@
...
+
+message VerificationMaterial {
+        oneof content {
+                // Signed with a raw key pair
+                PublicKey public_key = 1 [(google.api.field_behavior) = REQUIRED];
+                // Signed with cert + rfc3161 timestamp
+                Cert cert = 2 [(google.api.field_behavior) = REQUIRED];
+                // Logged to Rekor V3, authenticating with an ID token or
+                // a public key/signature.
+                TlogProof tlog_proof = 3 [(google.api.field_behavior) = REQUIRED];
```
Hayden-IO, line 93: Flagging for discussion: In the current v1 bundle, you can record a signature to multiple logs. In this new proposal, since we're equating a log proof to a signature, we allow only a single log proof.

If we wanted to continue to allow multiple log records, it would mean bundles can contain multiple 'signatures'. But then that'd be strange to allow multiple transparent signatures but not multiple PKI or managed key signatures.

protos/sigstore_bundle_v2.proto
Leont: I did wonder the same.

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,105 @@
...
+        TimestampVerificationData timestamps = 3 [(google.api.field_behavior) = REQUIRED];
+}
+
+message VerificationMaterial {
+        oneof content {
+                // Signed with a raw key pair
+                PublicKey public_key = 1 [(google.api.field_behavior) = REQUIRED];
+                // Signed with cert + rfc3161 timestamp
+                Cert cert = 2 [(google.api.field_behavior) = REQUIRED];
+                // Logged to Rekor V3, authenticating with an ID token or
```
Leont, line 91: So in the Identity information is now only in the transparency log body? (in some yet to be specified format?)

protos/sigstore_bundle_v2.proto
kommendorkapten: I think keeping a numerical identifier is the preferred option. As the entire bundle changes beyond the actual PQC changes, it's more like "Sigstore V2" but with only PQC algorithms.

protos/sigstore_bundle_v2.proto
kommendorkapten: Which doc are you referring too? The values that should go `extra` ought to be the real provenance values from the ID token: https://github.com/C2SP/C2SP/blob/main/tlog-proof.md#format I.e in the tlog proof.

protos/sigstore_bundle_v2.proto
kommendorkapten: Yes, that is correct. There are some discussion of the format, but I don't think it's finalized yet.

protos/sigstore_bundle_v2.proto
kommendorkapten: Why not use ASN.1, how would the client send the root of trust in the case of signing prior to submission? I feel like we can end up with something under specified if we don't specify how to identify public keys?

protos/sigstore_bundle_v2.proto
kommendorkapten: Agreed. Forcing a single tlog proof makes the bundle leaner, and so the verification logic. _If_ we ever need a signature threshold I'm inclined to think of a higher level construct in the client SDKs, and say ship a `.jsonl` file instead with multiple bundles.

protos/sigstore_bundle_v2.proto
Leont: IME a self-signed certificate is almost always a better idea than a bare public key for this sort of thing. A certificate actually has a name (and other metadata such as a subject key identifier), which can be used for this kind of matching.

protos/sigstore_bundle_v2.proto
Hayden-IO: SGTM, just wanted to confirm this was an explicit choice.

protos/sigstore_bundle_v2.proto
Hayden-IO: Correct. This is sketched out in https://github.com/sigstore/rekor-tiles/pull/830. The identity information will be a mapping between provenance name and value. Each key/value with be separated by `:`, and each line separated by `\n`. Should make parsing much easier, not having to deal with x509.

So for example, an identity token for a developer will result in a proof containing the following:

```
identity:hayden@foo.com
issuer:https://accounts.google.com
```

and for identity tokens from CI providers, it will use key names based on the OIDs from https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md:

```
identity:https://github.com/org/repo/ref/main@workflow.yml
issuer:https://token.githubactions.com
build_signer_digest:foobar
source_config_uri:....
```

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,105 @@
...
+        dev.sigstore.common.v1.HashOutput message_digest = 1 [(google.api.field_behavior) = REQUIRED];
+        // When signing, data is the artifact.
+        // MUST be unset when submitting with an ID token to a log.
+        optional Signature signature = 2;
+}
+
+message TlogProof {
+        // The log proof, formatted per c2sp.org/tlog-proof.
+        // If OIDC based submit is used, the `extra` field contains the
+        // relevant claims needed for recreation of the leaf during
```
Leont, line 54: I don't really like the sound of this "recreation" thing. I have bad experiences with that sort of thing (SETs being a good example). This will need some thorough documentation.

protos/sigstore_bundle_v2.proto
```diff
@@ -0,0 +1,105 @@
...
+        dev.sigstore.common.v1.PublicKeyIdentifier public_key = 1 [(google.api.field_behavior) = REQUIRED];
+        // Optional timestamp to enable key rotation when a signature is not
+        // logged (witnesses provides a counter-signature with timestamp).
+        optional TimestampVerificationData timestamps = 2;
+}
+
+message Cert {
+        oneof cert_data {
+                dev.sigstore.common.v1.X509Certificate certificate = 1 [(google.api.field_behavior) = REQUIRED];
+                dev.sigstore.common.v1.X509CertificateChain certificate_chain = 2 [(google.api.field_behavior) = REQUIRED];
```
Leont: Why have both certificate and certificate_chain? Isn't a certificate equivalent to a chain of length 1?

Leont: If we do want a chain, it might make more sense to split it into a `signing_certificate` field, and an `intermediate_certificates` repeated field. Any user will need to split the chain anyway.

protos/sigstore_bundle_v2.proto
Hayden-IO: Root of trust could be sent as an encoded public key, both Ed25519 and ML-DSA RFCs specify an encoding without needing PKIX/ASN.1. You could imagine something like `"ML-DSA-44" || 1312-byte pub key`. It's a simpler alternative to PKIX, which would provide the same information.

The ask to not use ASN.1 is because Sigsum proofs are verified on embedded hardware and they want to minimize heavier dependencies, ala ASN.1. I think self-signed certificates would get the same pushback.

protos/sigstore_bundle_v2.proto
Hayden-IO: We'll make sure the leaf is fully specified (https://github.com/C2SP/C2SP/pull/244) and document how to reconstruct the leaf hash. The client _must_ reconstruct the leaf hash when verifying the inclusion proof, which is what we meant by recreation.

With the current leaf types, unfortunately reconstructing the leaf hash required JSON canonicalization. Since some ecosystems lacked implementations, we also returned the canonicalized entry and let the client verify each component of the entry (e.g. the signature, artifact hash, public key/cert). One of the goals with the new leaf type is that canonicalization must be well-defined and straightforward.

And just to mention, there will no longer be SETs, as an inclusion proof and a timestamped checkpoint provides a guarantee of log inclusion.

protos/sigstore_bundle_v2.proto
Hayden-IO: I like that suggestion!

An alternative is to only have `signing_certificate`, and mandate that clients provide the chain during verification. But we know a few organizations that want to distribute intermediates in the bundle, because their intermediates are not long-lived (https://github.com/sigstore/sigstore-conformance/issues/342).

This is the one part of the bundle I feel less confident about, as we're bringing forward behavior from the previous bundle format that has led to some confusion around client implementations. Also with the current bundle format, I think only the Go SDK supports verifying with bundle-distributed intermediates, all other clients don't. Whatever decision we make, we'll want to drive this through conformance.

As an aside, it's also worth considering if we want to enforce ordering of the intermediate certs, or just have it as a cert pool (https://github.com/sigstore/protobuf-specs/issues/249) and let the x509 verifier build the chain.

protos/sigstore_bundle_v2.proto
Hayden-IO: Sorry, doc was the PQC design doc. Originally, I had written down that the values in `extra` would be hashes of provenance values, not the provenance values themselves. But if we want the tlog proof to double as a provenance document, we must include the original provenance values. That's what I went ahead and implemented, and the design doc now says the same.

protos/sigstore_bundle_v2.proto
kommendorkapten: The reason we kept cert and cert chain as separate was that it was to make it harder introduce bugs when verifying. We used to have one field for both but added a separate field. With a combined field you would need to configure the verifier to apply certain lengths checks which can lead to more complex code over reading a separate field and so have different logic for each option (cert or cert chain).

protos/sigstore_bundle_v2.proto
Leont: > The ask to not use ASN.1 is because Sigsum proofs are verified on embedded hardware and they want to minimize heavier dependencies, ala ASN.1.

That sounds like an [XY problem](https://en.wikipedia.org/wiki/XY_problem). You don't have to implement all of ASN.1 to implement SubjectPublicKeyInfo, it's quite trivial compared to the cryptography we're doing; in the end it's just another tag-length-value encoding.

> I think self-signed certificates would get the same pushback.

While certainly more complicated than public keys, that too can be much simpler than you might think. E.g. BearSSL has an embedded implementation that is simple enough to be malloc-free.

protos/sigstore_bundle_v2.proto
Leont: Any specific reason for not using JSON for this too? Might even be easier, and wouldn't break if anyone would ever put a newline in their data.

protos/sigstore_bundle_v2.proto
Hayden-IO: What's being suggested would still keep the fields separate. Either:

* We have `certificate` or `certificate_chain`, or
* `certificate`, and optionally `intermediates` to supplement.

In either case, whether the client should use the latter would need to be explicitly configured. I like the idea of having a dedicated `certificate` field that is always set, and moving away from one_ofs.

kommendorkapten: Yes, my bad. I agree, that is cleaner and with an `intermediates` field it becomes clearer too that we should not have any root certs. I'll update.

protos/sigstore_bundle_v2.proto
Leont: Apparently this sort of issue has hit sigstore before? https://github.com/sigstore/sigstore-go/pull/424#issuecomment-2720718126

kommendorkapten: To keep things simpler, we should avoid any constructs that can lead to situations like padding algorithms, or mixing hash/curve (the example you linked). 

This update specifies the exact application layer hashing to perform prior to computing the signature, which provides no option for user confusion (this is true for both ML-DSA and Ed25519 which are the only two algorithms currently in scope). The spec will clarify hedged/empty context etc). If we need any updates in the future due to a new PQC algorithm being added or a new mode, a new release of this repo has to be published with necessary changes to clarify.

If we think this is likely to happen, we can add a new reserved and unused field to avoid changing the wire format (but I rather not until there is a real need).

PR #907, opened 2026-07-09
Add monitor configuration protobuf definition

Hayden-IO:
This introduces a monitoring configuration, to be used by rekor-monitor and rekor-watch. This file will be shipped by TUF so that monitors of the public deployment can discover new log shards dynamically.

The trusted root is not sufficient, because the base URL for a tlog service is not guaranteed to be the read path. Additionally, a monitor needs to know the log version to construct the full path to entries, joining the signing config (which has the version) and the trusted root. This was possible in the past, but we recently changed the URL for the public log deployment to be a global LB, meaning the SC and TR can no longer be joined.

This also cleans up an outdated RFC 9162 reference (spec is not used) and adds some more language around how the TR URL for a log can be used.


protos/sigstore_trustroot.proto
```diff
@@ -26,14 +26,14 @@ option ruby_package = "Sigstore::TrustRoot::V1";
 
 // TransparencyLogInstance describes the immutable parameters from a
 // transparency log.
-// See https://www.rfc-editor.org/rfc/rfc9162.html#name-log-parameters
-// for more details.
 // The included parameters are the minimal set required to identify a log,
 // and verify an inclusion proof/promise.
 message TransparencyLogInstance {
         // The base URL at which can be used to URLs for the client.
```
cmurphy: could clean up this comment
```suggestion
        // The base URL that a client may use to identify a log.
```

Hayden-IO: Updated!

cmurphy (approved):

kommendorkapten (approved):

ret2libc (approved):

PR #936, opened 2026-09-02
move to grpc-web

bobcallaway:
drops a massive number of js/ts deps


kommendorkapten (approved):
`+19 -2,052` LOVE IT!

